PR #5, opened 2023-08-17
919 basic dataset handler

nimshi89:
This PR contains the first unit tests and the hello_app has been chnaged


src/hello_app.py
```diff
@@ -1,13 +1,21 @@
 #this script set up the basic flask app
```
mikeAdamss, line 1: we can lose this comment now.

also, can you rename the file to "main.py" and update the readme instructions accordingly please.

you'll also need to update the readme telling other developers how to run the tests.

src/hello_app.py
```diff
@@ -1,13 +1,21 @@
...
 app = Flask(__name__)
 
-@app.route("/", methods =["GET"])
-def main():
-    return "<h1 style='Blue:red'>Hello World!</h1>"
+@app.route("/datasets", methods =["GET"])
+def datasets():
+    if request.headers['Accept'] == "application/ld+json":
+        return get_json(request)
+    else:
```
mikeAdamss, line 11: you can just `return "", 406` in place of the else and lose the "not_recognised" function.

we need a "get_json" function ready for when things get complicated later, in the case of it not being a supported accept header we'll always just return a 406, no need to defer that to a function.

tests/test_hello_world.py
```diff
@@ -0,0 +1,41 @@
...
+    headers = {"Accept":"application/ld+json"}
+
+    # specify your end point
+    response = client.get('/datasets', headers=headers)
+    
+    # Assertions etc
+    assert response.status_code == 200
+
+
+def test_hello_world_handler_for_some_behaviour_fail():
```
mikeAdamss, line 25: rename to `test_unsupported_accept_headers_return_406` please

tests/test_hello_world.py
```diff
@@ -0,0 +1,41 @@
+import pytest
+
+from hello_app import app
+
+
+def test_hello_world_handler_correct_header_input():
```
mikeAdamss, line 6: rename to `test_jsonld_accept_header_returns_200` please

tests/test_hello_world.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    test the hello_world handler to conform its 
+    returning the expected values based on specific input
```
mikeAdamss, line 9: "Confirm that a 200 status code is returned where a "accept: application/json+ld" header is provided."

tests/test_hello_world.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+
+    # Create a test client
```
mikeAdamss, line 12: can you remove all the one line the comments in both tests please, that was me explaining, we don't need them in the codebase.

tests/test_hello_world.py
```diff
@@ -0,0 +1,41 @@
...
+    response = client.get('/datasets', headers=headers)
+    
+    # Assertions etc
+    assert response.status_code == 200
+
+
+def test_hello_world_handler_for_some_behaviour_fail():
+    """
+    test the hello_world handler to conform it fails
+    when a not specified data type is provided and returns the expected values
```
mikeAdamss, line 28: comment should be: 

"Confirm that a 406 status code is returned where an unsupported accept header is provided."

mikeAdamss (changes requested):
code is fine 👍 but some tidying up to do as per comments.

can you get rid of the pycache files as well please, they shouldn't be in git.

also ... I cant remember if you need the `./tests/__init__` here, not a biggie but can you try it without and remove it if its superfluous.

mikeAdamss (changes requested):
almost there 😄 

## deleting stuff
you need to delete the folders `./src/__pycache__` and `./src/__pycache__` and commit that deletion to git.

its probably also worth adding a `.gitignore` in the **base** of the repo and setting it to ignore `__pycache__` directories (literally just put the word `__pycache__` on the first line of an otherwise blank file named `.gitignore`), git will ignore all such directories in the future.

remember to commit that new `.gitignore` file as well.

can you also please delete `./tests/__init__.py` doesn't do anything in this context.
## filename

missed this earlier but `test_hello_world.py` doesnt make sense anymore. Make it `test_datasets.py` please.

## readme

Add some instructions on how to run tests please, just so it's obvious for folks in the future. `pipenv run pytest` from the command line plus a few lines on how to run tests via the vscode thing you've setup.

mikeAdamss (approved):
Approve, but make the following minor changes before you merge please.

Should be test_dataset**s**.py (plural) as thats the name of the endpoint.

Also dont specify "thiss will run the `test_dataset.py` file and run all tests in it" as what tests we've got will keep changing. Just saying how to run the tests is enough.

mikeAdamss (approved):

PR #7, opened 2023-08-22
923 adding database client

nimshi89:
Added a abstract class with a child class containing the get_dataset()


src/store/base.py
```diff
@@ -0,0 +1,56 @@
...
+    @abstractmethod
+    def setup(self):
+        ...
+
+    @abstractmethod
+    def get_datasets(self):
+        ...
+
+
+class SparqlStore(BaseStore):
```
mikeAdamss, line 21: can you put this in a different python file at the same level please.

i.e
- ./store/base.py - has BaseStore
- ./store/sparql.py - has SparqlStore

will make our lives easier when we start adding more store client later.

src/store/base.py
```diff
@@ -0,0 +1,56 @@
...
+
+    #seting up the self.url
+    def setup(self):
+        # set the url for sparql
+        self.url = os.getenv("SPARQL_ENDPOINT_URL")
+
+        #to make sure self.url is not none providin a default url
+        if self.url == None:
```
mikeAdamss, line 29: can see where you're coming from but don't set a default. Just assert its not None (with a helpful message saying what the issue is).

the idea is you _want_ your apps to refuse start when they're not configured correctly - then loudly shout about precisely what the issue is. Defaults can be non obvious and obvious is your friend when it comes to maintaining things.

src/main.py
```diff
@@ -1,4 +1,6 @@
 from flask import Flask, request
+from abc import ABC, abstractmethod
```
mikeAdamss, line 2: you dont need this here 😄

src/store/base.py
```diff
@@ -0,0 +1,56 @@
...
+            self.url = "https://beta.gss-data.org.uk/sparql"
+
+
+    def get_datasets(self) -> QueryResult:
```
mikeAdamss, line 33: so there's a point on restructuring here and a point on return types.

### restructuring

- 1.) We want to defined the _query string_ in the get_datasets() function, nice ✅ 
- 2.) We want _everything that uses the query string_ elsewhere in `self.run_sparql()`, ready for when we have many queries to run.

as an example (I've added some extra functions to get the idea across:

```

    def run_sparql(self, query):
        """ Runs and returns the results from a sparql query"""
        sparql = SPARQLWrapper(self.url)

        sparql.setQuery(query)
        sparql.setReturnFormat(JSON)
        return sparql.query()

    def get_datasets(self):
        """
        Get many datasets
        """
        query = """the sparql query to get many datasets"""
        result =  self.run_sparql(query)
        
    def get_datasets(self):
        """
        Get a single datasets
        """
        query = """the sparql query for getting one dataset"""
        result = self.run_sparql(query)
        
     def get_editions(self):
        """
        Get many editions
        """
        query = """the sparql query to get many editions"""
        result = self.run_sparql(query)
        
    def get_edition(self):
        """
        Get a single edition
        """
        query = """the sparql query for getting one edition"""
        result = self.run_sparql(query) 
```

does that makes more sense? With the current pattern we'll end up repeating ourselves quite a lot.

### return types

cant be QueryResult 😄 . The idea here is that _all_ implementations of a store have the same methods, take the same arguments and return the same thing - in this instance QueryResult is SPARQL specific.

think of it this way, if your next task was to create a SqlStore as an option we could use instead of SparqlStore then SqlStore.get_datasets() wont ever be returning a SPARQLWrapper.QueryResult, the thing being returned _has to be generic_.

Some sort of sensible dictionary structure is fine for now (we'll be updating the query in due course I'm sure), but make sure it is a dictionary and not a package specific class please.

tests/test_sparql_store.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+
+from store.base import SparqlStore
+from SPARQLWrapper import QueryResult
+
+
+def test_get_datasets():
```
mikeAdamss, line 7: you can make this `test_run sparql` and just test the method I've asked you to write this way.

you'll need a second test the to check that `get_datasets()` is returning a nice standard dict.

mikeAdamss (changes requested):
few requested changes, nothing major, you've got the hard bits done 👍

src/store/base.py
eldeal: Yes agree with everything Mike said here, my bad on not catching these/explaining when we chatted the other day

mikeAdamss (changes requested):
added a few extra comments, if you implement them then `get_datasets()` will be returning a json dict which is all we need for this task.

Pipfile
```diff
@@ -6,6 +6,7 @@ name = "pypi"
 [packages]
 flask = "*"
 gunicorn = "*"
+sparqlwrapper = "*"
```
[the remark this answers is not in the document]
nimshi89: Yeah sorry I did run the `pipenv install SPARQLWrapper` but I did change it manually, yeah I didn't know there is a difference sorry :)

tests/test_sparql_store.py
```diff
@@ -0,0 +1,15 @@
+from typing import Dict
+import pytest
+
+from store.sparql import SparqlStore
+
+
+def test_run_sparql():
```
mikeAdamss, line 7: you'll need a few more tests here:

- 1.) "test_sparql_store_setup_default" - test that when you instantiate a SparqlStore with no env var the default url is set
- 2.) "test_sparql_store_setup_custom" -  test that you can specify a different url via an env var and the SparqlStore will use that instead.
- 3.) "test_sparql_store_run_sparql" needs to test the _run_sparql_ method and confirm its returning a SPARQLWrapper.QueryResult.
- 4.) "test_sparql_store_get_datasets" needs to test the _get_datasets_ method and confirm its returning a dict.

mikeAdamss (changes requested):
almost there, just added some requests around test coverage.

mikeAdamss (changes requested):
one last tweak 😄

tests/test_sparql_store.py
```diff
@@ -0,0 +1,53 @@
...
+
+    assert test_instance.sparql.endpoint == "https://beta.gss-data.org.uk/sparql"
+    
+def test_sparql_store_setup_custom():
+    """
+    testing the setup function having the correct 
+    url when provided a custom env variable
+    """
+
+    os.environ["SPARQL_ENDPOINT_URL"] = "https://beta.gss-data.org.uk/custom_endpoint"
```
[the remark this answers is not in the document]
mikeAdamss: bit more of a clarification on this point from a slack chat:

...... 

so say I'm working on the api and I set an environment variable of `SPARQL_ENDPOINT_URL=www.mikesnewsparwlendpoint/sparql` , right now as soon as I run the tests it'll be replaced and deleted, so my code will start doing unexpected things (using the default endpoint even though _as far as I'm aware_ I've set a custom sparql endpoint on my machine via the env var).

so you just need to be careful not to change anything on the developers local machine when you run tests, there's two scenarios here really:

1.) Developer **doesn't** already have a `SPARQL_ENDPOINT_URL` set. In which case you're fine to do what you did, set it then delete it when the tests done :white_check_mark:
2.) Developer **does** already have a `SPARQL_ENDPOINT_URL` set on their machine, in which case it needs to still be set (to whatever it was originally) after your test has finished.

you need to account for both.

for (2) you'll need to (a) put the value of the old/existing env var in a python variable (b) change the env var so you can run your test (c) create the store (d) use the python variable to set the env var back to what it was (e) run the rest of your test.

mikeAdamss (approved):
LGTM

PR #8, opened 2023-08-31
924-structure-dataset

nimshi89:


src/store/sparql.py
```diff
@@ -1,8 +1,12 @@
 
 from store.base import BaseStore
```
mikeAdamss, line 2: so python imports are split into three blocks by convention.

```
<anything from the standard library>

<anything from a third party package>

<anything local>
```

And should be alphabetical within those blocks. So your imports should look like:

```
import os
from typing import Dict

from SPARQLWrapper import SPARQLWrapper, QueryResult

from store.base import BaseStore
```

src/store/sparql.py
```diff
@@ -13,10 +17,31 @@ def setup(self):
     def run_sparql(self, query) -> QueryResult:
         """ Runs and returns the results from a sparql query"""
         self.sparql.setQuery(query)
-        self.sparql.setReturnFormat(JSON)
```
mikeAdamss, line 16: did JSON not work? Always use a constant where one is available to you, strings can be typo'd, constants can't.

src/store/sparql.py
```diff
@@ -13,10 +17,31 @@ def setup(self):
...
+        self.sparql.setReturnFormat("json")
         return self.sparql.query()
     
-    def get_datasets(self):
+    def map_query_response_to_json(self, list_of_data):
+        nicer_list = []
+        for item in list_of_data:
+            n = {
+                "title": get_value_from_dict(item, "name"),
```
mikeAdamss, line 27: so this only says "name" because we specified ?name in the SPARQL query. It's probably worth specifying ?title in the SPARQL query instead so the mapping is more intuitive. i.e this could then be `"title": get_value_from_dict(item, "title"),`

src/store/sparql.py
```diff
@@ -13,10 +17,31 @@ def setup(self):
...
+                "description": get_value_from_dict(item, "description"),
+                "summary": get_value_from_dict(item, "comment"),
+                "last_updated": get_value_from_dict(item, "modified"),
+                "links": {"self": {"url": "Mike will get the value from Flask"},
+                "publisher": {"url": get_value_from_dict(item, "creator"),
+                              "id": get_value_from_dict(item, "creatorName")},
+                "topic": {"url": get_value_from_dict(item, "theme"),
+                          "id": get_value_from_dict(item, "themeName")},
+                "releases":{"url": get_value_from_dict(item, "comment")},
```
mikeAdamss, line 36: as discussed, you'll need to check in with @eldeal on this. "releases" and "latest_version" as these distinctions are not (as far as I know) implemented on the main PMD yet.

for the moment every dataset will be its own latest version so you can just use the id for the dataset you're looking at (see other comment) and the same "url" you're building for self.

"releases" I don't know, you'll need to check in with Eleanor

mikeAdamss (commented):
good progress on a hard thing here 👍 . Have left what I'm hoping is some helpful comments.

@eldeal - I'm unsure on the dataset ID here, I'm guessing we're gonna need that returned? Also not sure what it'd be or if there's a suitable pithy id we can use (ids in PMD tend to be urls). Might need some convos with data folks, unsure.

src/store/sparql.py
nimshi89: it in the dataset that we are suing the title is stored under name so that is how it is accessed

src/store/sparql.py
eldeal: Im surprised this isn't something an IDE or linter would automatically correct, is this maybe something we need to look into as a team standard to make sure we each have configured? But by all means yes lets follow the conventions even if by hand for now

src/store/sparql.py
eldeal: agree, latest version is the same as self for now. but yeah we can remove releases, we're far from that existing

eldeal (changes requested):
Good progress, only a few more tweaks I think

src/store/sparql.py
mikeAdamss: you dont need optional, in the sparql query the "?name" part is variable assignment. its us deciding we want to call the variable being returned ?name - you can just use ?title instead is all i mean.

as to why bother - longer term we want a future where data engineers or scientists provide the queries and us devs just use the results - so we want as much of the mapping of thing called A to thing called B in the query as possible, keeps the python code more inuitive (and tbh they're better placed to decide those mappings than us - areas of expertise n all).

src/store/sparql.py
```diff
@@ -1,13 +1,21 @@
...
+    """this function will aquire a value from a given list of dictionaries"""
+    
+    return item[name]["value"]
 
 class SparqlStore(BaseStore):
 
     #seting up the self.url
     def setup(self):
         url = os.environ.get("SPARQL_ENDPOINT_URL", "https://beta.gss-data.org.uk/sparql")
+        self.host = os.environ.get("HOST", "local host and the port it is running on")
```
mikeAdamss, line 18: "local host and the port it is running on" should literally just be "http:localhost:5000". That way when we run this without the HOST env var being set (i.e we run it locally) the urls will resolve correctly.

mikeAdamss:
putting this on hold as we've had a change in direction. May or may not close in time but keep the example code around for a little bit while we figure that out.

do not merge or work on.

mikeAdamss:
closing as we're done a pivot.

PR #9, opened 2023-09-18
Datasets handler storage

Moasib-Arif:


eldeal (approved):
Looks good to me, but probably worth just giving Jo a heads up before we merge this into his working branch. One to discuss at standup

Pipfile
```diff
@@ -6,8 +6,9 @@ name = "pypi"
 [packages]
 flask = "*"
 gunicorn = "*"
-SPARQLWrapper = "*"
 jsonschema = "*"
+sparqlwrapper = "*"
+pytest = "*"
```
mikeAdamss, line 11: we don't need this here. We probabaly need to update the readme but the way it works is:

- `pipenv install` - installs the libraries in packages
- `pipenv install -d` (d for dev) installs the libraries in "packages" **and** the libraries in "dev-packages".

You can just use the latter and remove this duplicate pytest entry from "packages".

It's a separation principle, we want "packages" to only contain libraries used when running in production, keeps things nice and clean.

Could you likewise move jsonschema down to dev-packages as well (and confirm the tests run ok please).

Pipfile
```diff
@@ -6,8 +6,9 @@ name = "pypi"
 [packages]
 flask = "*"
 gunicorn = "*"
-SPARQLWrapper = "*"
 jsonschema = "*"
+sparqlwrapper = "*"
```
mikeAdamss, line 10: Can you keep with SPARQLWrapper please. I know it's odd but we had differing behaviour of some kind between "sparqlwrapper" and "SPARQLWrapper" on the other branch.

Could have been something else going on (they seem the same on the surface) but if we change to the lower case install it'd be better to do it after it all gets merged together (better safe than sorry).

src/main.py
```diff
@@ -1,17 +1,28 @@
-from flask import Flask, request
-
+from flask import Flask, request, Response , json 
+from store.sparql import SparqlStore
 
 app = Flask(__name__)
+storage = SparqlStore()
+jsonld = "application/ld+json"
```
mikeAdamss, line 6: I like the use of a constant (good thinking) but:

- python contants are uppercase, `JSONLD` please.
- given this constant doesn't naturally sit in either file that's using it can you make a `constants.py` at the same level as `main.py` and import it from there please. Nice and obvious what it is then and encourages future folks to follow the same pattern with other mimetypes.

tests/test_datasets.py
```diff
@@ -16,6 +16,21 @@ def test_jsonld_accept_header_returns_200():
...
+    data = json.loads(response.data.decode("utf-8"))
+    assert len(data) > 0
+
+
+def test_empty_dataset_returns_404(client):
+    """
+    Confirm that a 404 status code is returned when the dataset is empty.
+    Assume storage.get_datasets() returns an empty dictionary for this test
+    """
+    headers = {"Accept": "application/ld+json"}
```
mikeAdamss, line 29: use your constant  😄

src/main.py
```diff
@@ -1,17 +1,28 @@
...
-    if request.headers['Accept'] == "application/ld+json":
-        return get_json(request)
+    status_code = 200
+    if request.headers['Accept'] == jsonld: 
+        result = storage.get_datasets()
+        
+        if len(result["items"]) == 0:
+            status_Code  = 404
+        
+        resp = Response(json.dumps(result))
```
mikeAdamss, line 17: there's a flask "jsonify" function for returning json objects, probably worth a quick investigation: https://www.geeksforgeeks.org/use-jsonify-instead-of-json-dumps-in-flask/

mikeAdamss (commented):
this is looking pretty grand - though I added a few extra comments while catching up, let me know if you have any questions 👍

mikeAdamss:
putting this on hold as we've had a change in direction. May or may not close in time but keep the example code around for a little bit while we figure that out.

do not merge or work on.

mikeAdamss:
closing as we've done a pivot.

PR #24, opened 2023-09-28
Creating StubMetadataStore class

nimshi89:


src/main.py
```diff
@@ -1,9 +1,12 @@
+
+import csv
 import os
 
+from pathlib import Path
 from fastapi import FastAPI, Request, Response, status
 
 from constants import JSONLD
-from store import StubStore
+from store import StubStore, StubMetadataStore
```
mikeAdamss, line 9: slight naming mixup.

the origin `StubStore` needs to be renamed to `StubMetadataStore` (and update anywhere it's referenced.

your new store needs to be `StubCsvStore`.

src/store/stub/store.py
```diff
@@ -1,4 +1,11 @@
 import json
+import csv
+from pathlib import Path
+from typing import List
+
+from fastapi.responses import FileResponse
+from store.base import BaseStore, BaseCsvStore
```
mikeAdamss, line 7: imports go:

```
default python library

third party libraries

local imports
```

so you're missing a gap, it should be

```
import json
import csv
from pathlib import Path
from typing import List

from fastapi.responses import FileResponse

from store.base import BaseStore, BaseCsvStore
```

src/store/stub/store.py
```diff
@@ -24,16 +31,16 @@ def setup(self):
             self.publishers = json.load(f)
 
         with open(Path(content_dir / "topics.json").absolute()) as f:
-            self.topics = json.load(f)
+            self.themes = json.load(f)
```
mikeAdamss, line 34: can you change themes to topics please, I had to correct that earlier on another branch so the history is well confusing, but its topics that we want.

src/store/stub/store.py
```diff
@@ -24,16 +31,16 @@ def setup(self):
...
 
-    def get_dataset(self, id: str) -> List[dict]:
```
mikeAdamss, line 29: go with get_dataset()l get_topic() and get_publisher() please (no mention of id in the method name anymore). This is another one I changed on a different pr.

tests/test_csv.py
```diff
@@ -0,0 +1,23 @@
+
```
mikeAdamss, line 1: for tests:

- you **dont** need this.
- we **do** need a test_version.py, you should be able to base it on the other tests and just test to make sure that the store.get_version() method is being called once.

src/store/stub/store.py
```diff
@@ -42,4 +49,29 @@ def get_publishers(self) -> dict:
...
+    A stub of a store that returns representative csv from
+    files stored on disk.
+    """
+    def setup(self):
+        self.datasets = {}
+
+        content_dir = Path(Path(__file__).parent / "csv/cipd/2022-3")
+
+        with open(Path(content_dir / "dataset.csv").absolute()) as f:
+            self.datasets["dataset"] = csv.reader(f)
```
mikeAdamss, line 66: so the field ("dataset") in this case needs to be matched by what you're passing in on line 75, so this should be 
```
self.dataset["cipd/2022-3/dataset"] = csv.reader(f)
```

do also not the comment on naming of therse csv files as well please.

src/main.py
```diff
@@ -48,7 +54,13 @@ def dataset_by_id(request: Request, response: Response, id: str):
     else:
         response.status_code = status.HTTP_406_NOT_ACCEPTABLE
         return
-    
+
+@app.get("/datasets/{dataset_id}/editions/{edition_id}/versions/{version_id}")
+def version(dataset_id: str, edition_id: str, version_id: str):
```
mikeAdamss, line 59: can you add the default `request: Request, response: Response` args please. We dont need them yet but we will need them, best to keep the pattern going.

mikeAdamss (commented):
awfully close now, nice one 👍 

Other than the comments, can you rename the csv to simple numbers please, 1.csv, 2.csv etc. That's more representative of what'll they'll be once we have real data.

mikeAdamss:
closed in favour of: https://github.com/GSS-Cogs/idpd-api-poc/pull/27

same code but this one has unverified commits so I migrated it.

PR #46, opened 2023-10-09
Added Schema for Edition/Editions endpoint

Moasib-Arif:


tests/fixtures/editions.py
```diff
@@ -0,0 +1,15 @@
+import json
+import pathlib
+import pytest
+
+@pytest.fixture
```
mikeAdamss, line 5: this wont work as the pytest.fixture decorator is already on the thing that's calling this. You can remove this.

tests/fixtures/editions.py
```diff
@@ -0,0 +1,15 @@
...
+def edition_data():
+    file_path = pathlib.Path("src/store/metadata/stub/content/editions.json")
+    with open(file_path, 'r') as json_file:
+        return json.load(json_file)["items"][0]
+    
+@pytest.fixture
```
mikeAdamss, line 11: this wont work as the pytest.fixture decorator is already on the thing that's calling this. You can remove this.

tests/test_editions.py
```diff
@@ -16,31 +19,58 @@
...
     response = client.get(ENDPOINT, headers={"Accept": JSONLD})
 
-    # Assertions
     assert response.status_code == status.HTTP_200_OK
     mock_metadata_store.get_editions.assert_called_once()
 
 
+def test_editions_invalid_structure_raises():
+    """
+    Confirm that the store.get_editions() method is
```
mikeAdamss, line 54: Can you write this docstring in the buller pointed style please. Makes it a fair bit easier to follow along with what the tesst is trying to accomplish.

tests/test_edition.py
```diff
@@ -16,31 +21,61 @@
...
     assert response.status_code == status.HTTP_200_OK
     mock_metadata_store.get_edition.assert_called_once()
 
 
+
+def test_edition_invalid_structure_raises():
+    """
+    Confirm that:
+    
+    the store.get_edition(id) method is
```
mikeAdamss, line 60: Can you write this docstring in the buller pointed style please. Makes it a fair bit easier to follow along with what the tesst is trying to accomplish.

mikeAdamss (commented):
this is excellent, only real issue is you forgot to `make test` before pushing, you'd have seen the issue straight away... **always** run tests before pushing, the one time you dont they'll be broken 😄 

other than that minor comments only.

mikeAdamss (commented):
so moving the fixture decorator into the fixtures directory 100% makes sense (good shout, cleaner code).

its a better apprach but comes with some minor tidy ups needed, can we:

- apply the same "decorators in the fixtures directory" approach for datasets as we are for editions please, need to keep things consistant.
- update the naming slightly, "response" is contextual (we might not be using this fixture to test an endpoint) and "expected" is implict. Can we go with the following please:
      `dataset_test_data`
      `datasets_test_data`
      `edition_test_data`
      `editions_test_data`
- can we add docstrings for the editions fixtures as well please, same style as the datasets ones.

otherwise this is looking grand, good job. get those tweaks it and we'll get this merged.

_edit_ - "test" is just asn implict as "expected" isn't it😆 /sigh ... just pick one of them and go with it, lose response though.

mikeAdamss (approved):
LGTM 👍

mikeAdamss (approved):

PR #47, opened 2023-10-10
/datasets/{id}/editions/{id}/versions endpoint#33 | Mock Data

Moasib-Arif:


tests/fixtures/versions.py
```diff
@@ -0,0 +1,12 @@
+import json
+import pathlib
+
+
+def versions_data():
+    """
+    Returns a dictionary representing the dictionary
+    we'd expect returned from store.get_datasets().
```
mikeAdamss, line 8: well done on this, you just need to update the docsting to store.get_versions().

tests/test_versions.py
```diff
@@ -0,0 +1,120 @@
...
+
+from fastapi import status
+from fastapi.exceptions import ResponseValidationError
+from fastapi.testclient import TestClient
+import pytest
+
+from constants import JSONLD
+from main import app, StubMetadataStore
+from fixtures.versions import versions_data
+# Devnotes:
```
mikeAdamss, line 11: blank line before #Devnotes for readability please.

src/store/metadata/stub/content/versions.json
```diff
@@ -0,0 +1,16 @@
+{
```
mikeAdamss, line 1: you'll need to add "in_dataset" and "in_edition" fields to go with your store.get_versions() holding logic.

ideally, we want to implement as far as we can, so building on the existing example we'll want to be hitting `/dataets/cpid/editions/2022-01` to trigger logic to return the contents of this file.

so in_dataset = "cpih" and in_edition="2022-01"

I'd also add one more mocked up item with a slightly different in_edition value just to make sure we're _only_ returning what's asked for.

src/store/metadata/stub/store.py
```diff
@@ -65,7 +65,17 @@ def get_edition(self, dataset_id: str, edition_id: str) -> Dict:
         return edition
 
     def get_versions(self, dataset_id: str, edition_id: str) -> Dict:
-        raise NotImplementedError
+        versions_for_edition = [
```
mikeAdamss, line 68: so if you look at the setup() method for the stub store we populate self.datasets, self.editions etc. You need to populate a self.versions there as well.

from there (once you've addressed the other comments) you should get a working endpoint.

mikeAdamss (commented):
this is good work 👍 

some minor things to finish wiring this up, but it doesn't need much.

mikeAdamss (approved):
LGTM

PR #48, opened 2023-10-10
Oxigraph

mikeAdamss:
## What is this

A combination of the following tickets:
- https://app.zenhub.com/workspaces/cogs-dd-cms-61e83b63053eb70011d37b4f/issues/gh/gss-cogs/idpd-api-poc/18
- https://app.zenhub.com/workspaces/cogs-dd-cms-61e83b63053eb70011d37b4f/issues/gh/gss-cogs/idpd-api-poc/22
- https://app.zenhub.com/workspaces/cogs-dd-cms-61e83b63053eb70011d37b4f/issues/gh/gss-cogs/idpd-api-poc/28

simply put - setup local oxigraph, populate oxigraph and use oxigraph for an endpoint. I needed to do these together to make sure it all came together at the end.

In terms of concrete things, this adds:
- a docker-compose and readme instructions for running your own local oxigraph.
- a script in /devdata for populating oxigraph from our jsonld dataset examples.
- an `OxigraphMetadataStore` with a working `get_dataset()` method that uses said data.
- a docker test runner that'll let us confirm that the above is working.

## How to review

- Follow the readme instructions and you should be able to get a local oxigraph running and populated with data. The example query from the readme should also work.
- `make test` should also work.
- to try the /datasets/cpih endpoint using the graph rather than the stub:
   - populate a local graph as per the readme.
   - change `Depends(StubMetadataStore)` to `Depends(OxigraphMetadataStore)` [here](https://github.com/GSS-Cogs/idpd-api-poc/blob/635781ae020b88ebde16f212ffc86526cd626a00/src/main.py#L41).
   - `export LOCAL_BROWSE_API=true`
   - hit `http://localhost:8000/datasets/cpih` and you should get a graph powered dataset response.
   - If you want to be clever, change some field values [here](https://github.com/GSS-Cogs/idpd-api-poc/blob/oxigraph/src/store/metadata/stub/content/datasets.json), repopulate the graph and go again, the values returned should changed accordingly.

I have left a couple to "TODO" statments which I'l capture in follow up cards, this is already a bit of a monster.

I am using a local context.json for now for the jsonsld<>RDF conversions. This is 100% wrong but will also 100% do for now.

Beyond that any  comments appreciated.


devdata/create_devdata.py
```diff
@@ -0,0 +1,163 @@
...
+            return "<bnode:b%s>" % node
+        return _node_to_sparql(node)
+
+    configuration = (f"{oxigraph_url}/query", f"{oxigraph_url}/update")
+    db = Dataset(store=SPARQLUpdateStore(*configuration, node_to_sparql=skolemise))
+    db.parse(seed)
+
+
+if __name__ == "__main__":
+    oxigraph_url = os.getenv("GRAPH_DB_URL", None)
```
StRhys, line 162: When running `pipenv run python3 ./devdata/create_devdata.py` I was getting an issue around getting the "GRAPH_DB_URL" var, unless I missed a setup step.
`AssertionError: You need to specfiy the oxigraph url via the OXIGRAPH_URL env var.`
I replaced it temporally with "http://localhost:7878" just to test.

mikeAdamss: nah, I forgot to document a step, good spot, I'll add it in 👍

StRhys (approved):
LGTM

PR #53, opened 2023-10-11
Added Documentation

Moasib-Arif:


src/main.py
```diff
@@ -11,15 +12,35 @@
...
     request: Request,
     response: Response,
     metadata_store: StubMetadataStore = Depends(StubMetadataStore),
 ):
+    """
+    Retrieve a list of datasets.
+
+    This endpoint returns a list of datasets available in the system.
+
+    Parameters:
```
mikeAdamss, line 34: this is a good doc string if this was just a python script ...but... this will be appearing as documentation for an api, and "parameter" has a different meaning there, they're the optonal arguments after a url, example:

```
https://some-domain/stuff?im-a-parameter=foo
```

So just leave it at the simple english description please, in this instance:

```
Retrieve a list of datasets.
    This endpoint returns a list of datasets available in the system.
```

src/main.py
```diff
@@ -34,12 +55,27 @@ def datasets(
...
 
 @app.get("/datasets/{dataset_id}", response_model=Optional[schemas.Dataset])
-def dataset(
+def get_dataset_by_id(
     request: Request,
     response: Response,
     dataset_id: str,
     metadata_store: StubMetadataStore = Depends(StubMetadataStore),
 ):
+    """
```
mikeAdamss, line 64: see comment on other endpoint

src/main.py
```diff
@@ -53,12 +89,27 @@ def dataset(
...
 @app.get("/datasets/{dataset_id}/editions")
-def editions(
+def get_dataset_by_editions(
     request: Request,
     response: Response,
     dataset_id: str,
     metadata_store: StubMetadataStore = Depends(StubMetadataStore),
 ):
+    """
+    Retrieve a list of editions for a specific dataset.
```
mikeAdamss, line 99: see comment on other endpoint

src/main.py
```diff
@@ -72,13 +123,29 @@ def editions(
...
-def edition(
+def get_dataset_edition_by_id(
     request: Request,
     response: Response,
     dataset_id: str,
     edition_id: str,
     metadata_store: StubMetadataStore = Depends(StubMetadataStore),
 ):
+    """
+    Retrieve information about a specific edition of a dataset.
```
mikeAdamss, line 134: see comment on other endpoint

src/main.py
```diff
@@ -92,11 +159,25 @@ def edition(
...
 
 @app.get("/publishers")
-def publishers(
+def get_all_publishers(
     request: Request,
     response: Response,
     metadata_store: StubMetadataStore = Depends(StubMetadataStore),
 ):
+    """
+    Retrieve a list of publishers.
```
mikeAdamss, line 168: see comment on other endpoint

src/main.py
```diff
@@ -111,12 +192,27 @@ def publishers(
...
 @app.get("/publishers/{publisher_id}")
-def publisher(
+def get_publisher_by_id(
     request: Request,
     response: Response,
     publisher_id: str,
     metadata_store: StubMetadataStore = Depends(StubMetadataStore),
 ):
+    """
+    Retrieve information about a specific publisher by ID.
```
mikeAdamss, line 202: see comment on other endpoint

src/main.py
```diff
@@ -130,11 +226,25 @@ def publisher(
...
 
 @app.get("/topics")
-def topics(
+def get_all_topics(
     request: Request,
     response: Response,
     metadata_store: StubMetadataStore = Depends(StubMetadataStore),
 ):
+    """
+    Retrieve a list of topics.
```
mikeAdamss, line 235: see comment on other endpoint

src/main.py
```diff
@@ -149,12 +259,27 @@ def topics(
...
 @app.get("/topics/{topic_id}")
-def topic(
+def get_topic_by_id(
     request: Request,
     response: Response,
     topic_id: str,
     metadata_store: StubMetadataStore = Depends(StubMetadataStore),
 ):
+    """
+    Retrieve information about a specific topic by ID.
```
mikeAdamss, line 269: see comment on other endpoint

src/main.py
```diff
@@ -169,14 +294,31 @@ def topic(
...
+def get_dataset_edition_version_by_id(
     request: Request,
     response: Response,
     dataset_id: str,
     edition_id: str,
     version_id: str,
     csv_store: StubCsvStore = Depends(StubCsvStore),
 ):
+    """
+    Retrieve a specific version and edition of a dataset.
```
mikeAdamss, line 306: see comment on other endpoint

src/main.py
```diff
@@ -53,12 +89,27 @@ def dataset(
 
 
 @app.get("/datasets/{dataset_id}/editions")
-def editions(
+def get_dataset_by_editions(
```
mikeAdamss, line 92: get_dataset_edtions please

src/main.py
```diff
@@ -11,15 +12,35 @@
...
+app = FastAPI(
+    title="IDPD-API-POC",
+    description="A proof of concept API layer to manage content negotiation across IDPD resources and provide an abstraction to 3rd party data services. This API provides access to various metadata, including datasets, editions, publishers, and topics.",
+    version="0.0.1",
+)
 
 
 @app.get("/datasets", response_model=Optional[schemas.Datasets])
-def datasets(
+def get_all_datasets(
```
mikeAdamss, line 24: it wasn't you but we've got a rogue additional `response.status_code = status.HTTP_200_OK` in this function, can you take it out as part of this pr please, thanks

mikeAdamss (commented):
This is good stuff (minor points only) but it highlights something interesting that makes me think we're using fastapi incorrectly, or at very least sub optimally.

If you look at the api docs for /dataset/{id} it automatically picking up the possibility of a 422 Validation Error response (cool). but raises two questions I need you to dig into:

- why is not picking up the potential validation error for the plain /datasets endpoint?
- how can we get it to pick up our 404 responses and 406 responses in a similar manner?

take your time and dig into it please. If there's a pattern we can follow here to have the documentation magically appear with all the variants of response in it, that'd be _very_ nice.

src/main.py
mikeAdamss: try and avoid the word list as its a list (or anything that implies structure) as its a list embedded in a dict and even that might change in time. Broad strokes.

src/main.py
```diff
@@ -11,17 +13,51 @@
...
+app = FastAPI(
+    title="IDPD-API-POC",
+    description="A proof of concept API layer to manage content negotiation across IDPD resources and provide an abstraction to 3rd party data services. This API provides access to various metadata, including datasets, editions, publishers, and topics.",
+    version="0.0.1",
+)
 
 
-@app.get("/datasets", response_model=Optional[schemas.Datasets])
-def datasets(
+@app.get(
```
mikeAdamss, line 24: so the statement `Optional` means "the thing in brackets **or** a None/Null response". So if our response model is `Optional[schemas.Datasets]` then any given documented response should be specifically returning one of those two options.

by way of an example, written like this...

```
@app.get(
    "/datasets", 
    response_model=Optional[schemas.Datasets], 
    responses={
        status.HTTP_200_OK: {
            "description": "Successful response. Returns a list of datasets.",
            "model": schemas.Datasets,
        },
        status.HTTP_404_NOT_FOUND: {
            "description": "Not Found. No datasets found.",
            "model": None,
        },
        status.HTTP_406_NOT_ACCEPTABLE: {
            "description": "Not Acceptable. The requested format is not supported.",
            "model": None,
        },
    }
)
```

the responses match the response model info, at the moment it's a bit dy-syncced.

could you apply this same logic across the other endpoints as well please.

src/main.py
```diff
@@ -11,17 +13,51 @@
...
+)
+def get_all_datasets(
     request: Request,
     response: Response,
     metadata_store: StubMetadataStore = Depends(StubMetadataStore),
 ):
+    """
+    Retrieve all the datasets.
+    This endpoint returns information on datasets available in the system.
+
```
mikeAdamss, line 50: can we lose the doctring beyond this point please, _**if**_ this was just a python application the below is usful information so I get why we're adding it.

...but... response & request objects, and certainly a store are programmer abstractions, they don't make sense when presented in the openapi doc.

same for the other endpoints please, just the simple plain english description will be fine for now.

mikeAdamss (commented):
a few comments but looking good.

after you address the comments can you merge main into this as well, you'll have a few conflicts to resolve but needs doing, give me a shout if that gets spicey.

mikeAdamss (approved):
LGTM

PR #60, opened 2023-10-16
Example Structure Logging for two Endpoints | For Review

Moasib-Arif:


mikeAdamss (commented):
well figured out, there's a few things to work out still but we've cooking with gas here:

1.) we'll need to `pipenv install structlog` to add it to the dependencies.

2.) We like to log such that the strucured logs are in keeping with the [dp logging standards](https://github.com/ONSdigital/dp-standards/blob/main/LOGGING_STANDARDS.md). 
In a _perfect_ world we'd want a call to (for example) logger.debug("I'm a debugging stuff")` to (under the hood)
   - auto populate namespace, trace_id, span_id, http, auth (namespace is "idpd-api-poc", the rest can be None for now, we'll do it later).
   - shift the python specific data (threads etc) over into the "data" field.
   - populate "event" with whatever the string positional argument is (i.e "I'm a debugging stuff").
   - severity we should (somehow) be able to get from the calling function, i.e debug/warning/error. there's some info here but each one represents severity expressed as a number: https://blog.sentry.io/logging-in-python-a-developers-guide/#:~:text=Let's%20enumerate%20the%20different%20logging,code%20that%20works%20as%20intended.
   - errors should be the error info
   .... etc

Basically, try and map over as best we can to the standard dp structures for these things. So a really straight forward logging statement in python, ends up as a nice detailed structured log.

One tricky bit will be the "data" field needs to be that python stuff **and** whatever a user might pass in via `data=`, don't let one overwrite the other.

You'll need to play, but creating something like....

```
class DpLogger:
    """
    A wrapper to create structured logs DP style.
    """

    def debug(self, event, data: Dict = None):
        # Add defaults things.
        # Do any mapping of what's getting passed so it makes sense
        # as the DP structure.
        logging.debug(<args and kwargs re-origanised>)
```

...is ok, if that's what it takes (though don't assume you need to, there might be a cleaner way).

3.) I was gonna leave async until we have to, but if it's easier to add it now to match that logging implementation then lets bite the bullet. but all or nothig please. Its all either aysnc or not.

4.) We'll need to figure out how to test this, that's gonna be fun.

TLDR: go for it, but make it all async and line it up with the DP standards please.

src/main.py
```diff
@@ -34,12 +83,13 @@ def datasets(
...
 
 @app.get("/datasets/{dataset_id}", response_model=Optional[schemas.Dataset])
-def dataset(
+async def dataset(
     request: Request,
     response: Response,
     dataset_id: str,
     metadata_store: StubMetadataStore = Depends(StubMetadataStore),
 ):
+    logger.info(f"Received request for dataset with ID: {dataset_id}", dataset_id)
```
mikeAdamss, line 92: right idea but using the data field for application specific varibles, eg:

```
logger.info("Received request for dataset by id", data={"dataset_id": dataset_id})
```

the idea is that by recording variables like this with fields, when structured logs ends up in one great big log store (where they're searchable) you'll be able to search for logs dealing with a specifc (in this example) dataset id. i.e use fields by preference for variables as they're more searchable that interpolated strings.

I wont comment them all, but can you update the others in the same way please.

src/main.py
```diff
@@ -1,25 +1,74 @@
...
+    except Exception as exc:
+        response = Response(status_code=500)
+        structlog.get_logger("api.error").exception("Uncaught exception", exc_info=exc)
+        raise
+    finally:
+        process_time = time.perf_counter_ns() - start_time
+        status_code = response.status_code
+
+    log_message = "Request completed"
+    log_context = {
```
mikeAdamss, line 43: so the way dp logging standards work is that anything applicaiton specific goes into the "data" field and these logging context items are going into the root of the dict being logged. So you'll want _something_ like:

```
log_context = {
        "data": {
               "path": request.url.path,
               "method": request.method,
               "status_code": status_code,
               "process_time": process_time / 10 ** 9,  # Convert to seconds
    }
}
```

to get these fields at the correct level.

the _tricky_ bit here is to make sure that if I do eg `log.info("something happened", data={"foo": "bar"})` then the key:value of {"foo": "bar") gets  _added_ to the data dict rather than replacing it.

mikeAdamss (commented):
I really like this, and the default log messages based on response code is super handy. A few comments, mainly around log structuring, the approch itself seems to be working grand.

Can you also try creating a middelware.py file (root level, alongside main.py), importing app into it and attaching the http middleware there please. If it works it'll be cleaner and we can keep main.py just as a declaration of routes.

You'd probably best merge in main and deal with any conflicts too, again, always better to take the pain as early as possible.

On the whole this is a complex thing and looking good.

mikeAdamss (commented):
couple of things from the merge that need adressing then this ones good to go.

we could burn more brain cells on log structuring but we'll come back to it later instead. I'd rather get it in and make sure it works in our infrastructure before expanding too much time on that.

mikeAdamss (approved):
LGTM, nice one.

PR #61, opened 2023-10-17
Data structures

mikeAdamss:
Updates the stub api and all dependant logic and tests to use the agreed data structures.

## How to review

- Pull this branch
- `pipenv sync --dev`
- `export LOCAL_BROWSE_API=true` - this will allow browser traffic despite mimetypes and will replace all `staging.idpd.uk` values with `localhost:8000` to allow click through navigation.
- `make start`
- go to http://localhost:8000/datasets

have a play around, make sure it looks sane and that things resolve to sensible json representations.

**Note** - there's a few TODO's and commented out code sections here relating to the graph, these will be covered in following tasks, this is already too big (and the graphs not relevant to the "make it representative" goal here).


src/schemas.py
```diff
@@ -59,79 +59,109 @@ class Distribution(BaseModel):
...
-        pattern=r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(Z|[+-]\d{2}:\d{2})?$"
-    )
-    next_release: str = Field(
-        pattern=r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(Z|[+-]\d{2}:\d{2})?$",
-    )
     publisher: str
     creator: str
     contact_point: ContactPoint
-    theme: Union[str, List[str]]
+    topics: Union[str, List[str]]
```
canwaf, line 73: Should we revert to `theme`? I would like to see our internal stuff directly aligned to the namespace, predicates, and objects we're employing and only through `@context` or the front end do we relabel things. Otherwise It's going to get way too confusing.

src/schemas.py
```diff
@@ -59,79 +59,109 @@ class Distribution(BaseModel):
...
     licence: str
+    issued: str = Field(
+        pattern=r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(Z|[+-]\d{2}:\d{2})?$"
+    )
+    modified: str = Field(
+        pattern=r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(Z|[+-]\d{2}:\d{2})?$"
+    )
     spatial_coverage: str = Field(pattern=r"^[EJKLMNSW]{1}\d{8}$")
-    temporal_coverage: PeriodOfTime
+    temporal_coverage: str
```
canwaf, line 84: Is this temporary? `PeriodOfTime` is the correct way of doing things.

src/schemas.py
```diff
@@ -59,79 +59,109 @@ class Distribution(BaseModel):
...
-    )
     publisher: str
     creator: str
     contact_point: ContactPoint
-    theme: Union[str, List[str]]
+    topics: Union[str, List[str]]
     frequency: Frequency
     keywords: list[str]
     licence: str
+    issued: str = Field(
```
canwaf, line 77: It would be good to add a `datetime.datetime.fromisoformat(dt_str)` validator instead of accepting 3000-99-32.

```python
def datetime_valid(dt_str):
    try:
        datetime.fromisoformat(dt_str)
    except:
        return False
    return True
```

src/store/metadata/stub/content/datasets.json
```diff
@@ -1,22 +1,28 @@
...
             "title": "Consumer Price Inflation including owner occupiers' housing costs (CPIH)",
             "summary": "The Consumer Prices Index including owner occupiers' housing costs (CPIH) is a measure of inflation that includes costs associated with...",
             "description": "The Consumer Prices Index including owner occupiers' housing costs (CPIH) is a measure of inflation that includes costs associated with...",
-            "release_date": "2017-01-01T00:00:00",
+            "issued": "2017-01-01T00:00:00",
+            "modified": "2017-01-01T00:00:00",
             "next_release": "2019-04-19T09:30:00Z",
-            "publisher": "office-for-national-statistics",
-            "creator": "office-for-national-statistics",
+            "publisher": "https://staging.idpd.uk/publishers/office-for-national-statistics",
```
canwaf, line 17: Our `@base` will be set for many of these fields, like [publisher,creator] only requiring the kebab case for the entities pointing to the URL for them on the gov.uk website.

So I'd expect to see `office-for-national-statitics` here. @rossbowen

src/store/metadata/stub/content/editions/versions/cpih_2022-01.json
canwaf, line 1: I like this, I'll add this to the application-profile.

src/schemas.py
mikeAdamss: so yeah I'm with you on that....but.. at the moment the stubbed jsonld responses we're using to power the api are also our source of data for loading in. So it has to be topics 😩 

conceptually no issues with keeping internal more rdf-y, just cant do that until we've progressed a ways and are looking at proper _distinct_ source files and no stub trickery.

canwaf (commented):
Let me know if you need approval, none of my comments are deal breakers, just if we're rolling from this and only updating it for production ready, there are a few gotchyas I'd not want to have to support fixing in the future. I'm not approving because I'm not officially on this yet.

src/schemas.py
mikeAdamss: It's just working to the example of edition I've got, which at the moment is this: https://github.com/GSS-Cogs/idpd-api-poc/blob/aa94d12909eb478e6616e2e1f4fe3b34a3427401/src/store/metadata/stub/content/editions/cpih_2022-01.json#L38

which I freely admit I don't understand. No issues with changing it, but could give me the new values for that field to match PeriodOfTime, would it just need to match the temporal coverage at the dataset level? i.e this: https://github.com/GSS-Cogs/idpd-api-poc/blob/aa94d12909eb478e6616e2e1f4fe3b34a3427401/src/store/metadata/stub/content/datasets.json#L37

canwaf: Temporal coverage is a span of time, temporal resolution is the list of intervals recorded.

Temporal Coverage: 
using iso8601, `2000-01-01/2009-12-31` or using the Period of Time notation in JSON-LD

```json
"temporal_coverage": {
"@type": "PeriodOfTime",
"start": "2000-01-01"
"end": "2009-12-31"
}

Resolution would be both a list of iso8601 values like, `["P1Y", "P1Q"]` for a dataset which has annual and quarterly resolutions (regardless of start date).

canwaf (approved):

src/schemas.py
rossbowen: My fault! I provided an older example. 🙈

src/store/metadata/stub/content/datasets.json
rossbowen: We had some discussion yesterday about themes and publishers, and whether they should be IRIs or kebab'd strings here. I'm still thinking kebab'd strings, but we agreed to do IRIs for now and do a small check with some suppliers to see what they favoured.

PR #68, opened 2023-10-19
Cloud storage

nimshi89:
This is a PR for replacing the StubCsvStore (thing that downloads CSVS as stored in this repo) with a GoogleStorageCsvStore that downloads the file(s) in question from a google bucket.

It's a little gnarly as we need to support async streaming downloads ("chunking") as some of the files are and will be pretty large.

## How to review

Sanity check code and tests.

Spin up the server and try this: `curl http://localhost:8000/datasets/beis-2005-to-2019-local-authority-carbon-dioxide-co2-emissions/editions/2005-2019/versions/1.csv -H "accept: text/csv" -o ./output.csv` the file in question is 260MB, that curl command should stream it (gradually download it) to your terminal, this is the same mechanism that'll powers your browser gradually downloading a big thing.

You can also add an `or BROWSABLE` to [this line](https://github.com/GSS-Cogs/idpd-api-poc/blob/bd45f428cff742dfd712d9c1c4e64d6373f609b0/src/main.py#L318) and put the large url ending csv from the above paragraph into your browser.

Note: I did need to change the return of `csvstore.get_version()` to a tuple (from a dict in the stub store) to keep the clear separation between request handler and store. Given we'll be removing the csv stub store if this works (probably early post demo) then its fine, just fyi of why I did a thing we'd usually try and avoid.


mikeAdamss:
Looks like (superficially, didn't look for long) a nasty dependancy issue in here somewhere. If anyone wants to investigate before I get to it please do 😄

SarahJohnsonONS (approved):
LGTM (although not professing to understand every part of it). The `curl` command results in a csv being downloaded.

mikeAdamss:
nobody merge please, this code is fine, but exacerbates a problem elsewhere.

mikeAdamss:
closing this but keeping the branch (for now). We won't be switching to this any time soon.

PR #69, opened 2023-10-19
Added Schema

Moasib-Arif:


src/schemas.py
```diff
@@ -45,7 +45,7 @@ class PeriodOfTime(BaseModel):
 class Column(BaseModel):
     name: str
     datatype: str
-    titles: str
+    title: str
```
mikeAdamss, line 48: needs to be title**s** plural, this is 100% my fault as I seem to have changed it in versions.json when I shouldnt have, my bad. As it stands this change makes the version tests pass but makes the edition tests fail (as they both use `Column` and are currently different.

just change it to title**s** plural in the versions.json and this back as well and we're good - make sure `make test` passes though, just in case.

mikeAdamss (commented):
nice work, minor change thats my fault not yours.

while you're in here can you add two other minor things (save me writing tiny task cards).

can you add the `Versions` schema here: https://github.com/GSS-Cogs/idpd-api-poc/blob/c2d1ad72e597450e3343b406d06cb3310f364008/data.py#L103

and the `Editions` schema here: https://github.com/GSS-Cogs/idpd-api-poc/blob/c2d1ad72e597450e3343b406d06cb3310f364008/data.py#L90

its just validating the data coming in, you can see usage up a little from those for the `Datasets` schema. thanks.

data.py
```diff
@@ -87,7 +92,76 @@ def populate(oxigraph_url=None, write_to_db=True):
         editions_source_dict = json.load(f)
 
     # Validate then add to graph
-    # TODO - add schema validation
+    class ContactPoint(BaseModel):
```
mikeAdamss, line 95: @Moasib-Arif - not quite what I meant, I meant _use_ the `Versions` and `Editions` schemas (as imported via `from src import schemas`. See this example of using the `Datasets` schema: https://github.com/GSS-Cogs/idpd-api-poc/blob/89e99d7f0d3235bcb787b69cadc71229ac0eb7c3/data.py#L80

mikeAdamss (commented):
looks good but two points of tidy up we'll need to do:

- theres unused imports in data.py at the moment, if you `make fmt` if should tell you about them.
- we'll need to add the responses for the versions endpoint (and some docstrings) to bring them inline with the other endpoints please (the api documentatiopn patterns you established elsewhere).

mikeAdamss (approved):

PR #82, opened 2023-10-24
Load multiple editions and versions source jsonlds

Moasib-Arif:


data.py
```diff
@@ -82,26 +98,24 @@ def populate(oxigraph_url=None, write_to_db=True):
...
     # Load from disk
-    editions_source_path = Path(subbed_metadata_store_content_path / "editions/cpih_2022-01.json")
-    with open(editions_source_path) as f:
-        editions_source_dict = json.load(f)
+    editions_source_dict = subbed_metadata_store_content_path / "editions"
 
     # Validate then add to graph
     # TODO - add schema validation
-    g += Graph().parse(data=json.dumps(set_context(editions_source_dict)), format="json-ld")
+    schemas.Editions(**editions_source_dict)
```
mikeAdamss, line 105: so this won't work, at the point you run it `editions_source_dict` is still just a `Path` object not a dictionary.

given the pattern you're using with `process_json_files` I would pass the schema in as an argument so you can validate after we json load and attach context in that function.

data.py
```diff
@@ -25,6 +26,21 @@
 
 from src import schemas
 
+def process_json_files(g, dir_path, context_file):
```
mikeAdamss, line 29: right idea but you cant pass context file in, it only exists in the scope of the "set_context" function.

its not the most efficient, but you can just do `g += Graph().parse(data=json.dumps(set*context(resource_dict)), format="json-ld")` at the end of this for now.

data.py
```diff
@@ -153,4 +167,5 @@ def skolemise(node):
 
 if __name__ == "__main__":
     oxigraph_url = os.getenv("GRAPH_DB_URL", None)
+    context_file = Path("src/store/metadata/context.json")
```
mikeAdamss, line 170: I see what you were doing now, but this script isnt always called by main (if you look at the Makefile you'll see a whacky inline python way of calling). Try it yourself with `make data`.

either way we need setting context in its own function as that local context file is a temporary thing, its (very likely) going to be more complicated than that and I'd look it encapsulated in its own obvious "change me" place.

mikeAdamss (commented):
a few tweaks but we're largely there.

src/main.py
```diff
@@ -109,7 +109,11 @@ def get_dataset_by_id(
     responses={
         status.HTTP_200_OK: {
             "description": "Successful response. Returns all the editions for the dataset.",
+<<<<<<< HEAD
```
mikeAdamss, line 112: some left over merge kurft here you'll need to resolve.

I won't tag them all but I think this file (and test_version) slipped in.

mikeAdamss (approved):
LGTM

PR #85, opened 2023-10-27
Updated metadata

Moasib-Arif:
**DONT MERGE THIS EVEN IF APPROVED** ... there's a syning thing between this and the frontend. Feel free to review and comment just dont merge, we need to do that in sequence with Rhys. --Mike--


data.py
```diff
@@ -83,15 +83,14 @@ def populate(oxigraph_url=None, write_to_db=True):
...
-        datasets_source_dict = json.load(f)
-        # Validate then add to graph
-        schemas.Datasets(**datasets_source_dict)
-        g = Graph().parse(
-            data=json.dumps(set_context(datasets_source_dict, context_file)),
-            format="json-ld",
-        )
+    # Load from disk
+    datasets_source_path = Path(
+        subbed_metadata_store_content_path / "datasets"
```
mikeAdamss, line 88: this is off. If you think about it you're saying the _directory_ is `idpd-api-poc/src/store/metadata/stub/content/datasets` - so its looking for a json file in that non existent directory, not finding any (naturally) and falling through without doing anything.

Whereas we want to validate and load the _specific_ file `idpd-api-poc/src/store/metadata/stub/content/datasets.json.`

Right idea (reuse resuse) but its not quite right.

data.py
```diff
@@ -100,60 +99,51 @@ def populate(oxigraph_url=None, write_to_db=True):
...
     # ------------------
     # Topics resources
     # ------------------
 
-    topics_source_path = Path(subbed_metadata_store_content_path / "topics.json")
-    with open(topics_source_path) as f:
-        topics_source_dict = json.load(f)
-
+    topics_source_path = Path(
+        subbed_metadata_store_content_path / "topics"
```
mikeAdamss, line 127: see comment for datasets

data.py
```diff
@@ -100,60 +99,51 @@ def populate(oxigraph_url=None, write_to_db=True):
...
+    process_json_files(g, topics_source_path, context_file, schema)
 
 
     # --------------------
     # Publishers resources
     # --------------------
 
     publishers_source_path = Path(
-        subbed_metadata_store_content_path / "publishers.json"
+        subbed_metadata_store_content_path / "publishers"
```
mikeAdamss, line 139: see comment for datasets

mikeAdamss (commented):
to sanity check if the data and schemas match up use `make data`, at the moment I think we've got some errors in that area, something about time. Needs investigating.

mikeAdamss (changes requested):
toggling to requested changes for visibility

mikeAdamss (dismissed):

data.py
```diff
@@ -93,54 +93,49 @@ def populate(oxigraph_url=None, write_to_db=True):
     # TODO - need to iterate all files in ./edtions not just the one
```
mikeAdamss, line 93: can you remove the todo while you're in here please, we've done it.

data.py
```diff
@@ -93,54 +93,49 @@ def populate(oxigraph_url=None, write_to_db=True):
...
     # Validate then add to graph
     schema = schemas.Editions
-    process_json_files(g, editions_source_dict, schema)
+    process_json_files(g, editions_source_path, schema)
 
     # ------------------
     # Versions resources
     # ------------------
 
     # TODO - need to iterate all files in ./editions/versions not just the one
```
mikeAdamss, line 107: can you remove the todo while you're in here please, we've done it.

data.py
```diff
@@ -93,54 +93,49 @@ def populate(oxigraph_url=None, write_to_db=True):
...
 
     # ------------------
     # Topics resources
     # ------------------
 
-    topics_source_path = Path(subbed_metadata_store_content_path / "topics.json")
-    with open(topics_source_path) as f:
-        topics_source_dict = json.load(f)
-
+    topics_source_path = Path(
```
mikeAdamss, line 120: this pattern won't work for the individual files, its saying "look in a directory equal to /src/store/metadata/stub/content/topics ..." which isn't a directory so no json files are found. So `process_json_files` is running but not processing anything.

we need some assertions around this behaviour (silent fails are bad - so its on my list of cards to add) but for now can you make sure that the implementation for topic and publishers follows the pattern as used as main branch.

data.py
```diff
@@ -93,54 +93,49 @@ def populate(oxigraph_url=None, write_to_db=True):
...
+    schema = schemas.Topics
+    process_json_files(g, topics_source_path, schema)
 
     # --------------------
     # Publishers resources
     # --------------------
 
     publishers_source_path = Path(
-        subbed_metadata_store_content_path / "publishers.json"
+        subbed_metadata_store_content_path / "publishers"
```
mikeAdamss, line 132: see comment on Topics processing.

mikeAdamss (changes requested):
good work on the schemas.

there's a little tidy up from the merge needs doing but it wont take much.

mikeAdamss (dismissed):

mikeAdamss (approved):

PR #86, opened 2023-10-27
Bug fix csv download

mikeAdamss:
Ad hoc bug fix as its interfering with a frontend task.

The csv download functionality from the stub csv store wasnt quite working. Literally just a filename typo and a lack of test coverage to tell us about it before now.

Have:
- fixed
- added some logging so its more transparent if there's future issues
- added some simple tests
- did a little ad hoc cleanup while was in there.

## How to review

Sanity check the code changes.

You can try the file response via curl:

```
curl http://localhost:8000/datasets/cpih/editions/2022-01/versions/1.csv -H "Accept: text/csv"
```

-H sets the header so the handler know the type being requested, which (should) then get recognised by the handler so that it calls the store to download the file response which (as you're in a terminal) will dump the csv contents into said terminal. Should work with and without the file extension.


SarahJohnsonONS (approved):
All seems good to me - got the curl command to return the csv in the terminal. I had to change `from src.custom_logging import logger` to `from custom_logging import logger` in `store/csv/stub/store.py` to be able to debug `main.py` and get the csv terminal output.

SarahJohnsonONS (approved):

PR #89, opened 2023-10-30
minor data fixes

mikeAdamss:
logic has gotten slightly snarled in data.py (how we turn the jsonlds into ttl), just a consequence of many prs and merge conflict resolutions on the bounce (and because its not wrapped in tests.... yet).

problem was some calls to `set_context` that are missing a positional argument (if you try and `make data` on main you'll see).

this does a little tweaking and resolves that.

## how to review

- sanity check.
- confirm you can `make data` and it executes and creates `./out/seed.ttl`.

For some wider context:
- we have stubbed jsonld data in /src/store/metadata/stub/content
- `make data` (and `make populate` once the graph parts are progressed) confirms these jsonld files match the expected schema and converts them into the seed.ttl - this is what we'll use to initially populate the graph.
- `make data` stops short of pushing to oxigraph so is how we sanity check that the correct ttl is being made without having to have an oxigraph up and running.

Key point - the stubbed data we're using to make the api look real is also our initial data source for the graph (json-ld is a serialisation of RDF, why not).

### DIFF
--- a/data.py
+++ b/data.py
@@ -21,7 +21,7 @@ from rdflib.plugins.stores.sparqlstore import SPARQLUpdateStore, _node_to_sparql
 from src import schemas
 
 
-def process_json_files(g, dir_path, context_file, schema):
+def process_json_files(g, dir_path, schema):
     """
     Process JSON files in a directory, apply context, and add to an RDF graph.
     """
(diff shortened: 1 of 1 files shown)


Moasib-Arif (approved):
All seems to look good 👍

PR #97, opened 2023-11-02
Get publisher_id

nimshi89:
created a function to get the publisher id value


src/store/metadata/oxigraph/store.py
```diff
@@ -138,7 +139,29 @@ def get_publisher(self, publisher_id: str) -> Optional[Dict]:  # pragma: no cove
...
+
+        # Serialize the graph into jsonld
+        data = json.loads(result.serialize(format="json-ld"))
+
+        # Use a context file to shape our jsonld, removing long form references
+        data = jsonld.flatten(
+            data, {"@context": constants.CONTEXT, "@type": "dcat:publisher"}
+        )
+
+        # Use a remote context
```
SarahJohnsonONS, line 161: We don't need this remote context apparently

src/store/metadata/context.json
```diff
@@ -140,6 +140,7 @@
     },
     "publishers": {
         "@id": "dcat:publisher",
+        "@type": "dcat:publisher",
```
SarahJohnsonONS, line 143: This is weird - I can see that it's needed, because it breaks when I remove it, but I feel like there should be a better solution (don't know what that solution is though!)

SarahJohnsonONS (changes requested):
A few minor tweaks, but looks good to me otherwise.

tests/store/metadata/oxigraph/test_oxigraph_publisher.py
```diff
@@ -0,0 +1,38 @@
...
+     """
+     Confirm that the OxigrapgMetadataStore.get_dataset()
+     function returns a dataset that matches the dataset
+     schema.
+     """
+
+     store = OxigraphMetadataStore()
+
+     publisher = store.get_publisher("office-for-national-statistics")
+     schemas.Publisher(**publisher)
```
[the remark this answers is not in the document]
nimshi89: Sorry I thought if that `schemas.Publisher(**publisher)` goes well that would  be sufficient. Based on other tests but yeah I can make an assertion :)

SarahJohnsonONS (approved):

PR #98, opened 2023-11-02
OxigraphMetadataStore get_topic method

SarahJohnsonONS:
- `get_topic()` method added to `store/metadata/oxigraph/store.py`, and three tests (topic with subtopic, topic with parent topic and topic schema validation). Had to do some jiggery-pokery to get the right bit of the graph to populate the schema model due to the `@context` changes (see L177-178) - this probably needs to be addressed as Jo was having a similar problem.
- Added `default_factory=list` to `schema.Topic` as I was getting a `missing` Pydantic Validation Error (even though those fields are meant to be Optional *shrugemoji*).
- Three new SPARQL queries added to construct the RDF needed to populate the schema model.


src/store/metadata/oxigraph/store.py
```diff
@@ -145,11 +145,38 @@ def get_topics(self) -> Optional[Dict]:  # pragma: no cover
         """
         raise NotImplementedError
 
-    def get_topic(self) -> Optional[Dict]:  # pragma: no cover
+    def get_topic(self, topic_id: str) -> Optional[Dict]:  # pragma: no cover
```
mikeAdamss, line 148: you can remove the # pragma: no cover please. It's just there so we didnt feedback of a lack of test coverage for a method that wasn't implemented.

src/store/metadata/oxigraph/store.py
```diff
@@ -145,11 +145,38 @@ def get_topics(self) -> Optional[Dict]:  # pragma: no cover
...
+        # Serialize the graph into jsonld
+        data = json.loads(result.serialize(format="json-ld"))
+
+        # Use a context file to shape our jsonld, removing long form references
+        data = jsonld.flatten(
+            data, {"@context": constants.CONTEXT, "@type": "dcat:theme"}
+        )
+
+        # Use a remote context
+        data["@context"] = "https://data.ons.gov.uk/ns#"
```
mikeAdamss, line 174: this can go. If we _were_ using context it'd need to be against `result["@context"]`, but given the schemas need updating for a context (which we'll need, but not right now) that'll keep for another piece of work, you can just delete.

mikeAdamss (changes requested):
this is grand 👍  just a couple of comments to consider and all things that just haven't come up as of yet.

one less obvious one: we're returning `sub_topics: []` here, which is cool (probably better than null in fairness) but you'll need to change the "null" values in [here](https://github.com/GSS-Cogs/idpd-api-poc/blob/main/src/store/metadata/stub/content/topics.json) otherwise Rhys could build an interaction that works on the stub but breaks when we switch over (i.e he might check for a null).

need to merge main across as well, I suspect it's just formatting.

tests/store/metadata/oxigraph/test_oxigraph_topic.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_oxigraph_get_topic_with_subtopic_returns_valid_structure():
+    """
+    Confirm that the OxigraphMetadataStore.get_topic()
+    function returns a topic with sub_topic that matches the topic
+    schema.
+    """
+
+    os.environ["GRAPH_DB_URL"] = "http://localhost:7878"
```
mikeAdamss, line 17: so for this, you should be able to use a pytest monkeypatch and conftest.py to set this envionrment variable but only for the length of the test run (at the moment this works, but you're overwriting the users env var, they might have it set to something else for a reasons).

there's a comment down a bit [here](https://stackoverflow.com/questions/46733332/how-to-monkeypatch-the-environment-using-pytest-in-conftest-py) that on the surface looks about right, though you might need to play. I'd start with the "Direct use of instance:" bit and see if it works.

mikeAdamss (commented):
minor thing to investigate.

mikeAdamss (approved):
looks grand, good job 👍

mikeAdamss:
closing, changes merges as part of https://github.com/GSS-Cogs/idpd-api-poc/pull/107

PR #115, opened 2023-11-07
CPIH less stubby now

canwaf:
As promised for Mike, let me know what needs changing. Hopefully xsd:datetime and xsd:decimal are valid datatypes for you.


mikeAdamss (approved):
LGTM

PR #117, opened 2023-11-07
Update CPIH dataset with improved time series and

canwaf:
Updates to improve text descriptions, better date representation, clarity of contents, metadata, etc.


mikeAdamss:
closing as Ross did a similar one over here that we used for the demo: https://github.com/GSS-Cogs/idpd-api-poc/pull/157, so will just get that other one working.

I don't have a horse in this race, let me know if we want it the other way around.

Issue #56, opened 2023-10-11
implement oxigraph.get_edition() method


mikeAdamss:
## What is this

We need to create a method of`OxiGraphMetadataStore.get_edition()` that can construct a jsonld response of a structure (a) matching the `Edition` schema.

## What to do

You can model your approach on what we're going for `Oxigraph.get_dataset()`.

Be careful with variable names in the sparql, they'll need to match the fields you're aiming to populate otherwise the jsonld context trickery will not work as expected.

So **write the test first** (base it on the existing "test_oxigraph_dataset" test (its a simple one) and code to get it to pass.

A mismatched mapping in jsonld won't be obvious and can sometimes just result in an omission rather than an error - you'll want the test.

## Acceptance Criteria

- [ ] Test passese, returned data matches Edition schema when the /datasets/{id}/editions/{id} endpoint is using `OxigraphMetadataStore`.

Update from Nick 03/11
I didn't manage to get this working in time before leaving for the weekend, so I think someone else will need to pick it up while I am away. I pushed everything I did so far into branch "#56-implement-get-oxigraph-edition", hopefully what is there along with these notes will help.

I reorganized everything I had written so it matches the style we seem to be forming with these get functions, but started to run into trouble with SPARQL queries afterwards. I ran the queries in the oxigraph server and got expected results of subject "<https://staging.idpd.uk/datasets/cpih/editions/2022-01>" which was fine. So the construct_edition functions in construct.py were written based off those queries, but also using the existing get_dataset and the recently written get_topic queries/functions for reference on the approach and structure. After making said structural changes I started getting errors, but I think in the latest commit it is almost there apart from the fact my queries are now not retrieving three fields from the Edition schema: versions_url, versions, and table_schema.

The get_edition function itself was mostly written based off the get_dataset function and it definitely needs more work. But I believe what's there follows the existing approach the other functions are using. At the moment of writing I am getting the error "AttributeError: 'function' object has no attribute 'namespaces'" when running the test for this function. I suggest using debugging, taking a look at the Edition schema and comparing it to what results are being generated in the function, then seeing which step causes the function to fail and reworking the function to suit expectations.

The test "test_oxigraph_get_edition_returns_valid_structure" was cut down to the basics because I was having trouble debugging what I am getting from the sparql queries, so I am not sure I did it correctly with the input/approach I used on it. I started the ticket by writing it and I don't have a lot of confidence in its correctness now so feel free to discard it if needed.

PR #120, opened 2023-11-08
#56 implement oxigraph get edition

SarahJohnsonONS:
- `get_edition()` method added in `store/metadata/oxigraph/store.py`. Mostly self-explanatory, but populating the table_schema.columns might need a bit of finessing - there's almost certainly a better way of doing it (in an ideal world).
- SPARQL queries added:
  - `construct_edition_core`
  - `construct_edition_table_schema`
  - `construct_edition_versions`
  - `construct_edition_contact_point`
  - `construct_edition_themes` (will rename to `construct_edition_topics` for consistency)
  - `construct_edition_keywords`
  - `construct_edition_temporal_coverage`
- context.json amendments:
```
    "versions": {
        "@id": "hydra:member",
        "@type": "@id",
        "@container": "@set"
    },
    "versions_url": {
        "@id": "hydra:Collection",
        "@type": "@id"
    },
```
- Tests added - lots of assertions in `test_oxigraph_get_edition_returns_valid_structure()` - do we need to check all fields?
- I can't see where the `titles` field in schemas.Column is defined in the context, so I've commented it out in the schema model - will reinstate once it's been added to the context.


src/schemas.py
```diff
@@ -45,7 +45,7 @@ class PeriodOfTime(BaseModel):
 class Column(BaseModel):
     name: str
     datatype: str
-    titles: str
+    # titles: str
```
mikeAdamss, line 48: we'll need this, can see why you commented it out though, we're missing a context item for "titles" so its missing from the RDF, this should do it

```
    "titles": {
        "@id": "csvw:titles",
        "@language": "en"
    },
```

and just update your construct sparql to pull it in,

src/store/metadata/oxigraph/store.py
```diff
@@ -110,7 +117,73 @@ def get_edition(
...
+            (x for x in data["@graph"] if "dcat:endDate" in x.keys()), None
+        )
+        table_schema_graph = next(
+            (x for x in data["@graph"] if "columns" in x.keys()), None
+        )
+        columns_graph = [x for x in data["@graph"] if "datatype" in x.keys()]
+
+        versions_graph = next((x for x in data["@graph"]))
+        # Populate table_schema_graph.columns with column definitions and delete table_schema_graph.column blank node @id
+        for column in table_schema_graph["columns"]:
```
mikeAdamss, line 159: if I'm following you should be able to do something like:

```
for column in columns_graph:
      del column["@id"]
```

and just `edition_graph["table_schema"]["columns"] = columns_graph` later on

src/store/metadata/oxigraph/store.py
```diff
@@ -110,7 +117,73 @@ def get_edition(
...
+        edition_graph["temporal_coverage"] = {
+            "start": temporal_coverage_graph["dcat:startDate"]["@value"],
+            "end": temporal_coverage_graph["dcat:endDate"]["@value"],
+        }
+        # Populate editions_graph.table_schema.columns with column definitions and delete editions_graph.table_schema blank node @id
+        edition_graph["table_schema"]["columns"] = table_schema_graph["columns"]
+        del edition_graph["table_schema"]["@id"]
+
+        # Expand version to include issued and modified fields
+        for idx, version in enumerate(edition_graph["versions"]):
```
mikeAdamss, line 180: I would consider something like:

- get all graphs in @graph with an id of `/<something>/versions/<something>` into a list and `edition_graph["versions"] = [that list]`

you can probably create the list with a one line lambda tbh, eg something along the lines of:

`versions_graphs = [x for x in data["@graph"] if re.match("some appropriate regex", x["@id"])]`

[some stuff on regular expressions in python](https://www.w3schools.com/python/python_regex.asp) and otherwise just play here until you've got a decent expression: https://regexr.com/

note - my off the cuff syntax might not be 100%, but it'd be along those lines.

mikeAdamss (commented):
so dont shoot the messenger but this has unverified commits, you'll need to bounce it to a new branch again (and merge conflicts etc 😄) 

RE tests - no, you don't need all the assertions, the schema checks the required fields are there, as long as they are then our code doesnt care whats in them (if someone wants to use our api to put trash on the internet that's there perogative - provided its trash in the right structure). That said, a small handful of assertions is probably sensible just as a sanity check, focus them on anything we're explicitly munging  if you can.

Other than that a few comments that I hope are helpful

src/store/metadata/oxigraph/store.py
```diff
@@ -110,7 +117,73 @@ def get_edition(
...
+        # Serialize the graph into jsonld
+        data = json.loads(result.serialize(format="json-ld"))
+
+        # Use a context file to shape our jsonld, removing long form references
+        data = jsonld.flatten(
+            data, {"@context": constants.CONTEXT, "@type": "dcat:Dataset"}
+        )
+
+        edition_graph = next((x for x in data["@graph"] if "@type" in x.keys()), None)
+        contact_point_graph = next(
```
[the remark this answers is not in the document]
SarahJohnsonONS: Would it be worth including a couple of the previous steps, e.g. including an argument for the `data = json.loads(...)` step and for the `@type` key of the `ctx` dict that we pass to `flatten()`? DRY...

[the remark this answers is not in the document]
SarahJohnsonONS: Ignore this, realised it was a stupid idea

src/store/metadata/oxigraph/store.py
SarahJohnsonONS: I absolutely suck at regex, so I've done it like this, which I think should do the job, but let me know if not.

src/store/metadata/oxigraph/store.py
```diff
@@ -110,7 +118,60 @@ def get_edition(
...
+        data = json.loads(result.serialize(format="json-ld"))
+
+        # Use a context file to shape our jsonld, removing long form references
+        data = jsonld.flatten(
+            data, {"@context": constants.CONTEXT, "@type": "dcat:Dataset"}
+        )
+
+        edition_graph = _get_single_graph_for_field(data, "@type")
+        contact_point_graph = _get_single_graph_for_field(data, "vcard:fn")
+        temporal_coverage_graph = _get_single_graph_for_field(data, "dcat:endDate")
```
mikeAdamss, line 148:
```
if None in [edition_graph, contact_point_graph, temporal_coverage_graph]:
         return None
```

its not perfect (there's no guarantee the dicts include the keys we expect them to - though sparql being sparql thats unlikely to be a thing) but it'll guard us against uncontrolled exceptions if we try and reference a key inside a None further down the function

src/store/metadata/oxigraph/store.py
```diff
@@ -110,7 +118,60 @@ def get_edition(
...
+        edition_graph["contact_point"] = {
+            "name": contact_point_graph["vcard:fn"]["@value"],
+            "email": contact_point_graph["vcard:hasEmail"],
+        }
+        edition_graph["temporal_coverage"] = {
+            "start": temporal_coverage_graph["dcat:startDate"]["@value"],
+            "end": temporal_coverage_graph["dcat:endDate"]["@value"],
+        }
+
+        # Add `issued` and `modified` fields to each version in `versions`
```
mikeAdamss, line 166: comment needs a quick update (or delete if you'd prefer), its not really doing this anymore

src/store/metadata/oxigraph/store.py
```diff
@@ -110,7 +118,60 @@ def get_edition(
...
+        versions_graph = [
```
mikeAdamss, line 167: this is pendantic af and I wouldnt block a pr for this alone, but `version_graphs` not `versions_graph`

mikeAdamss (changes requested):
well done, couple of tiny changes only - though we'll to recommit this for the lack or early verified commits.

mikeAdamss:
closed in favour of: https://github.com/GSS-Cogs/idpd-api-poc/pull/127

PR #121, opened 2023-11-08
Regex fix

StRhys:
This is to fix a warning that was a appearing when running `make test`, with the warning being `DeprecationWarning: invalid escape sequence \.`


mikeAdamss (approved):
LGTM

PR #122, opened 2023-11-08
support multiple editions and versions in stubs

mikeAdamss:
addressing some TODO comments that have been around for a while - this PR allows the metadata and csv stub to populate itself more than one edition and version json.

it's literally just "glob all the files" rather than "read _the_ file".

## how to review

- confirm you can make start and make test and the expected things happen.
- sanity check


StRhys (approved):
LGTM

PR #125, opened 2023-11-09
get_publishers()

nimshi89:
adding the `get_publishers()` function


src/store/metadata/oxigraph/store.py
```diff
@@ -258,7 +259,40 @@ def get_publishers(self) -> Optional[Dict]:  # pragma: no cover
...
+        )
+
+        # Serialize the graph into jsonld
+        data = json.loads(result.serialize(format="json-ld"))
+
+        # Use a context file to shape our jsonld, removing long form references
+        data = jsonld.flatten(
+            data, {"@context": constants.CONTEXT, "@type": "hydra:Collection"}
+        )
+
```
SarahJohnsonONS, line 281: Add a comment here so we know to go back and remove this when the context weirdness is resolved

src/store/metadata/oxigraph/store.py
```diff
@@ -258,7 +259,40 @@ def get_publishers(self) -> Optional[Dict]:  # pragma: no cover
...
+        # TODO Fix context weirdness - at the moment, the flatten() method is changing @type to `versions_url`
+        # also adds dcat: to the publishers
+        data["@graph"][0]["@type"] = "hydra:Collection"
+        data["@graph"][0]["publishers"] = data["@graph"][0].pop("dcat:publisher")
+
+        for idx, publisher in enumerate(data["@graph"][0]["publishers"]):
```
SarahJohnsonONS, line 287: Mike said try and avoid using `enumerate` if you can use a comprehension instead, e.g. this works as well:
```
    data["@graph"][0]["publishers"] = [
        self.get_publisher(x["@id"].split("/")[-1])
        for x in data["@graph"]
        if "landing_page" in x.keys()
    ]
    del data["@graph"][0]["dcat:publisher"]
```

src/store/metadata/oxigraph/store.py
```diff
@@ -258,7 +259,40 @@ def get_publishers(self) -> Optional[Dict]:  # pragma: no cover
...
+            publisher_id = publisher["@id"].split("/")[-1]
+            data["@graph"][0]["publishers"][idx] = self.get_publisher(publisher_id)
+        
+        # TODO Update @context so it's not hardcoded
+        data["@graph"][0]["@context"] = "https://staging.idpd.uk/ns#"
+        result = data["@graph"][0]
```
SarahJohnsonONS, line 293: `@context` is still missing so the test is failing on validation

SarahJohnsonONS (commented):
Comments added

src/store/metadata/oxigraph/store.py
SarahJohnsonONS: Weird, this wasn't here when i added the comment

nimshi89: Yeah it is odd, I don't see why this issue remains but for now it solves it

PR #127, opened 2023-11-10
new branch for verification

SarahJohnsonONS:


src/store/metadata/oxigraph/store.py
```diff
@@ -102,15 +112,95 @@ def get_editions(self, dataset_id: str) -> Optional[Dict]:  # pragma: no cover
...
+        data = jsonld.flatten(
+            data, {"@context": constants.CONTEXT, "@type": "hydra:Collection"}
+        )
+        editions_graph = _get_single_graph_for_field(data, "@type")
+
+        # TODO Fix context weirdness - at the moment, the flatten() method is changing @type to `versions_url` and `editions` to `versions`
+        editions_graph["@type"] = "hydra:Collection"
+        editions_graph["editions"] = editions_graph.pop("versions")
+
+        for idx, edition in enumerate(editions_graph["editions"]):
```
mikeAdamss, line 137: you can use a comprehension for this

```
        editions_graph["editions"] = [
            self.get_edition(dataset_id, x["identifier"]) for x in editions_graph["editions"]
        ]
```

unless there's multiple nested conditions. you can almost always use a comprehension, handy string to have on your bow.

as a rule of thumb in python, use a comprehension until such time as it impacts readability, at which point (for the sake of the next developers sanity) you'll want to iterate.

src/store/metadata/oxigraph/store.py
```diff
@@ -102,15 +112,95 @@ def get_editions(self, dataset_id: str) -> Optional[Dict]:  # pragma: no cover
...
+        result: Graph = construct_editions(graph, dataset_id)
+
+        # Serialize the graph into jsonld
+        data = json.loads(result.serialize(format="json-ld"))
+
+        # Use a context file to shape our jsonld, removing long form references
+        data = jsonld.flatten(
+            data, {"@context": constants.CONTEXT, "@type": "hydra:Collection"}
+        )
+        editions_graph = _get_single_graph_for_field(data, "@type")
```
mikeAdamss, line 131: check if its None and return None if it is please, gotta guard against key errors later on.

src/schemas.py
```diff
@@ -61,6 +61,21 @@ class Distribution(BaseModel):
     table_schema: TableSchema
 
 
+class SummarisedVersion(BaseModel):
+    """
+    A short form schema for Version as presented
+    at the /datasets level
```
mikeAdamss, line 67: /editions level

tests/store/metadata/oxigraph/test_oxigraph_editions.py
```diff
@@ -0,0 +1,29 @@
...
+    schema.
+    """
+
+    store = OxigraphMetadataStore()
+
+    editions = store.get_editions("cpih")
+    editions_schema = schemas.Editions(**editions)
+    assert editions_schema.id == "https://staging.idpd.uk/datasets/cpih/editions"
+    assert editions_schema.type == "hydra:Collection"
+    assert len(editions_schema.editions) == 1
```
mikeAdamss, line 23: can we change this to assert len > 0 please. that way if we update that sample dataset with more editions later on we won't need to update the test.

mikeAdamss (changes requested):
this is looking good, couple of minor comments only.

mikeAdamss (approved):
LGTM

PR #129, opened 2023-11-10
Data merge

mikeAdamss:
amalagamted pr of all the DEs additional data work (with a few tweaks where needed).

have also (a) upped the character limit on summary and description to accommodate some of the more verbose stuff and (b) change ordering in main.py slightly (makes it easier to sanity check, otherwise no functional difference).

## how to review

mainly need a sanity check, lots of fiddly things to line up here, need another pair of eyes to make sure I haven't missed anything. You _dont_ need to check by reading raw json files, just check by using the api (as follows).

- export LOCAL_BROWSE_API=true
- make start
- click around, make sure you can navigate, all urls should resolve to a thing.
- make sure you can `make data` without errors.

you should. also be able to

```
curl <some url ending in .csv> -H "accept: text/csv" 
```

to get csv data.


StRhys (approved):
LGTM

PR #131, opened 2023-11-13
Context endpoint added

SarahJohnsonONS:
- `self.context` added to `StubMetadataStore.setup()` (currently hardcoded to get the local context.json file)
- `StubMetadataStore.get_context()` method added
- `"/"` endpoint defined in `main.py`
- Tests added for HTTP responses 200, 404 and 406
- Setting LOCAL_BROWSE_API=true and navigating to localhost:8000, the context is visible:
image


src/main.py
```diff
@@ -32,6 +34,39 @@
...
+    This endpoint returns information on the context.
+    """
+
+    logger.info("Received request for context")
+    if request.headers["Accept"] == JSONLD or BROWSABLE:
+        context = metadata_store.get_context()
+        if context is not None:
+            response.status_code = status.HTTP_200_OK
+            return context
+        response.status_code = status.HTTP_404_NOT_FOUND
```
mikeAdamss, line 63: can you log an error here please (still return a 404, we just want an error logged so we dev types know somethings not right).

for (most of, see next paragraph) the other endpoints a 404 is normal behaviour (they take variables, so someone asking for something that's not there is inevitable and expected). In the case of the context endpoint though, it _should_ always be there as there's no incoming variables, so if its not we'll want to know about it.

to save me raising another ticket could you add an error logging statement for /datasets, /publishers and /topics on the 404 response route as well please while you're in here, same logic.

src/store/metadata/stub/store.py
```diff
@@ -57,6 +57,7 @@ def setup(self):
         using the contents of ./content
         """
         content_dir = Path(Path(__file__).parent / "content")
+        context_dir = Path(Path(__file__).parent.parent)
```
mikeAdamss, line 60: so the code is good but I wouldn't put it in the store class.

so it works, but there's no Oxigraph version of context (a populated graph doesnt care about a jsonld context file) so we'd have to move the context logic somewhere ekse in order to delete the stub store (which we'll be doing relatively soon).

I would probably just add a context.py at the root level for now.

mikeAdamss (changes requested):
well done, minor comments only plus a slight path change coming out of a convo with Ross.

src/store/metadata/stub/store.py
SarahJohnsonONS: OK, this is a bit of a weird one - I tried moving `get_context()` to a `context.py` file and calling the function directly in `get_context()` in `main.py`, but I was getting a failure for `assert_called_once()` in the tests (i.e. nothing was being called). So I created a new class called `ContextStore` (currently in `base.py`), so that I could include it as a dependency, which fixed the problem. Let me know if you think there's a better way to do it.

src/store/metadata/base.py
```diff
@@ -86,3 +88,12 @@ def get_sub_topic(self, topic_id: str, sub_topic_id: str) -> Optional[Dict]:
         """
         Get a specific sub-topic for a specific topic
         """
+
+
+class ContextStore(ABC):
```
mikeAdamss, line 93: this is clever but we don't need to inherit ABC, thats just for enforcing that the class that extends the class has the `@abstractmethod` decorated methods, not applicable here.

also can we move it to another file, say "context.py" or similar, just to avoid the implication from being in base.py that its an abstract class.

mikeAdamss (changes requested):
looking good, one minor comment on structuring

mikeAdamss (approved):
LGTM

PR #136, opened 2023-11-14
moved changes from 58-get-version to this branch

MuazzamChaud:
adding get_version() and the test_oxigraph_get_version_returns_valid_structure(), found at /tests/store/metadata/oxigraph/test_oxigraph_version.py should pass

### DIFF
--- a/tests/store/metadata/oxigraph/test_oxigraph_version.py
+++ b/tests/store/metadata/oxigraph/test_oxigraph_version.py
@@ -1,0 +1,43 @@
+import os
+import pytest
+
+from pydantic import ValidationError
+
+from store.metadata.oxigraph.store import OxigraphMetadataStore
+import schemas
+
+
+def test_oxigraph_get_version_returns_valid_structure():
+    """
+    Confirm that the OxigraphMetadataStore.get_version()
+    function returns a version with table schema that matches the version
+    schema.
+    """
+    store = OxigraphMetadataStore()
+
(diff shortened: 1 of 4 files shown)


tests/store/metadata/oxigraph/test_oxigraph_version.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_oxigraph_get_version_returns_valid_structure():
+    """
+    Confirm that the OxigraphMetadataStore.get_version()
+    function returns a version with table schema that matches the version
+    schema.
+    """
+    store = OxigraphMetadataStore()
+
+    version = store.get_version(dataset_id="cpih", edition_id="2022-01", version_id="1")
```
mikeAdamss, line 18: Don't treat a positional arguments as keyword arguments please, should be:

```
store.get_version("cpih", "2022-01", "1")
```

mikeAdamss (commented):
Well done, one minor comment and a missing test and we're done.

RE missing test - on the other store method tests we're also confirming the schema logic is working. There's no particular reason it needs to be in the oxigraph test files (its on the refactor list) but while it is we need to be consistent.

Either:
- (a) add the same "I break" test to your new test file
- (b) create a lower level test_schemas.py and add the test there (and copy all of the other similar ones over).

dealers choice, we just cant ommit it where its included elsewhere as we want the codebase to be consistent.

mikeAdamss (approved):
LGTM

PR #138, opened 2023-11-15
Stub subtopics endpoint added

SarahJohnsonONS:
- `get_sub_topics` endpoint added in `main.py`
- `get_sub_topics()` in `StubMetadataStore` reworked to get the correct format returned as json
- `sub_topic_test_data` added to `fixtures.topic` (with new content stub `subtopics.json`)
- 4 new tests added and passing in `test_sub_topics.py`


src/store/metadata/stub/content/subtopics.json
```diff
@@ -0,0 +1,19 @@
+{
```
mikeAdamss, line 1: we shouldn't need this file, as a sub topic is also a thing from the topic level (and because this is our source of truth for seed RDF) we'll end up loading the same thing into oxigraph twice, will get very confusing if the definitions diverge.

src/store/metadata/stub/store.py
```diff
@@ -178,15 +178,21 @@ def get_sub_topics(self, topic_id: str) -> Dict:
...
+            for x in topics["topics"]
+            if len(x["parent_topics"]) > 0 and topic["@id"] in x["parent_topics"]
+        ]
+
+        if len(sub_topics) == 0:
             return None
         return contextualise(
             {
-                "items": sub_topics,
+                "@id": "https://staging.idpd.uk/topics",
```
mikeAdamss, line 193: well done, but you'll need to update the id too, i.e `@id: "http://localhost:8000/topics/{topic_id}/subtopics",`

tests/fixtures/topics.py
```diff
@@ -23,3 +25,14 @@ def topics_test_data():
     file_path = pathlib.Path("src/store/metadata/stub/content/topics.json")
     with open(file_path, "r") as json_file:
         return json.load(json_file)
+
+
+@pytest.fixture
+def sub_topic_test_data():
```
mikeAdamss, line 31: so this is just for checking the structure is right and the structure is the same as for topics so just return that and use the doscstring to explain, something like:

```
def sub_topic_test_data():
     """
     explanation
     """
     return topics_test_data()
```

If pytest moans about fixtures calling other fixtures (it might, unsure) then pull out the logic in `topics_test_data` into a stand along function and have both fixtures call it.

mikeAdamss (changes requested):
this is looking good, couple of minor comments only.

mikeAdamss (approved):
LGTM

PR #139, opened 2023-11-15
context corretions

mikeAdamss:
The ContextStore was getting confused about where to file the json file (worked perfectly locally, didn't work on k8s as the app start up in a slightly different folder) so tweaked the logic slightly.

Other than that its just correcting all the places I wrote the context url down incorrectly 😄


nimshi89 (approved):
looks fine to me

Issue #134, opened 2023-11-13
implement oxigraph.get_subtopics(topic_id) method


mikeAdamss:
## What is this

We need to implement a `.get_subtopics(topic_id)` method in the oxigraph store.

Unlike the other plural methods, there's no singular version of this (a child of a. topic is itself just a topic, just presented hierarchically).

## What do to

Create the method in line with the other. using the same patterns we've established so far. Also write tests along the established lines.

## Acceptance Criteria

- [ ] Endpoint Works
- [ ] Tests pass

PR #141, opened 2023-11-16
#134 implement oxigraph.get subtopics(topic id) method

SarahJohnsonONS:
Minly self-explanatory I think - there's a bit of wrangling going on in `get_sub_topics()` in `OxigraphMetadataStore`, so if anyone can figure out if some of that could be done in SPARQL, please let me know.


src/store/metadata/oxigraph/store.py
```diff
@@ -344,7 +348,46 @@ def get_sub_topics(self, topic_id: str) -> Optional[Dict]:  # pragma: no cover
         """
         Get all sub-topics for a specific topic
         """
-        raise NotImplementedError
+
+        # Populate the graph from the database
```
mikeAdamss, line 352: there are times in every deveopers career where you feel the need to kick yourself hard with a big boot 😄 

your code works (and well figured out btw) ....but... you can probably just do this sorta thing:

```
    def get_sub_topics(self, topic_id: str) -> Optional[Dict]:  # pragma: no cover
        """
        Get all sub-topics for a specific topic
        """

        all_topics = self.get_topics()
        topics_with_parents = [topic for topic in all_topics["topics"] if topic.get("parent_topics", None) is not None]
        sub_topics = [
            topic for topic in topics_with_parents if any([x.endswith(topic_id) for x in topic["parent_topics"]]) 
        ]
        
        if len(sub_topics) == 0:
            return None

        return {
            "@context": "https://staging.idpd.uk/ns#",
            "@id": f"https://staging.idpd.uk/topics/{topic_id}/subtopics",
            "@type": "hydra:Collection",
            "topics": sub_topics,
            "count": {len(sub_topics)},
            "offset": 0
            }
```

mikeAdamss (commented):
all works but I've put an alternative solution in a comment, we can do this a bit simpler 😄

src/store/metadata/oxigraph/store.py
SarahJohnsonONS: Thanks, I knew there was an easier way to do it *tunnel vision from looking at it for too long*

mikeAdamss (approved):
LGTM

PR #143, opened 2023-11-16
remove holding text

mikeAdamss:
Just getting rid of the holding text fields so Rhys has something representative to work with.

This is **not** final (not my bag) that will go over the data teams for finessing, this is just a "more representative holding data so we can build things" PR, i.e dont worry if Mikes descriptions are dodgy, that's fine, its better than a TODO.

Also added the coverage file to the gitignore so it'll go away.


Moasib-Arif (approved):
LGTM

PR #145, opened 2023-11-17
70 get versions

MuazzamChaud:
Added get_versions() and the test_oxigraph_versions.py should pass. Also added a change in get_version(), to change some instances of version_graph["issued"] from a List[str], to the right str value from said List[str].


src/store/metadata/oxigraph/store.py
```diff
@@ -252,6 +284,16 @@ def get_version(
         version_graph["table_schema"]["columns"] = columns_graph
         del version_graph["table_schema"]["@id"]
 
+        # if version_graph["issued"] is a list that contains ["2023-11-06T12:25:28", "2023-11-06T12:25:28+00:00"],
```
mikeAdamss, line 287: So some good figuring out here but I suspect this is a data error that has since been corrected (or its a data error local to your machine). If ever you hit unexpected data like this locally check the data on https://staging.idpd.uk/sparql and see if your local graph differs.

To clear our you local data:

- docker-compose down
- delete the `.data` folder

Either way we shouldn't be correcting incorrect data, making sure its correct is someone else's job.

At most check that this is a string as expected, if its _not_ then log an error so we know about it and return None.

src/store/metadata/oxigraph/store.py
```diff
@@ -219,6 +220,37 @@ def get_versions(
...
+        data = jsonld.flatten(
+            data, {"@context": constants.CONTEXT, "@type": "hydra:Collection"}
+        )
+        versions_graph = _get_single_graph_for_field(data, "@type")
+        if versions_graph is None:
+            return None
+
+        # TODO Fix context weirdness - at the moment, the flatten() method is changing @type to `versions_url` and `editions` to `versions`
+        versions_graph["@type"] = "hydra:Collection"
+        versions_graph["versions"] = versions_graph.pop("versions")
```
mikeAdamss, line 245: this isn't doing anything, we're effectively renaming "versions" to "versions".

the code you were basing this off has an issue where flatten was naming the "editions" fields as "versions" which doesnt apply here - you can remove the TODO for the same reason.

tests/store/metadata/oxigraph/test_oxigraph_versions.py
```diff
@@ -0,0 +1,30 @@
...
+    store = OxigraphMetadataStore()
+
+    versions = store.get_versions("4gc", "2023-09")
+    versions_schema = schemas.Versions(**versions)
+    assert versions_schema.context == "https://staging.idpd.uk/ns#"
+    assert versions_schema.id == "https://staging.idpd.uk/datasets/4gc/editions/2023-09/versions"
+    assert versions_schema.type == "hydra:Collection"
+    assert versions_schema.title == "4G Coverage"
+    assert len(versions_schema.versions) > 0
+    assert versions_schema.count == 1
```
mikeAdamss, line 24: best not check this, the tests will start failing if someone adds a version 2. You could assert it matches len(versions_schema.versions) if you like.

mikeAdamss (changes requested):
well done, minor comments only

src/store/metadata/oxigraph/store.py
mikeAdamss: Just clicked this is what @SarahJohnsonONS was telling me about earlier, worth checking in with her and seeing if this is corrected on the data side.

src/schemas.py
```diff
@@ -183,7 +183,7 @@ class Version(BaseModel):
     type: List[str] = Field(alias="@type")
     id: str = Field(alias="@id")
     identifier: str
-    issued: str = Field(
+    issued: Optional[str] = Field(
```
mikeAdamss, line 186: You don't need the optional, its either correct or we dont return it at all.

src/store/metadata/oxigraph/store.py
```diff
@@ -260,6 +263,36 @@ def get_versions(self, dataset_id: str, edition_id: str) -> Optional[Dict]:
...
+        # Use a context file to shape our jsonld, removing long form references
+        data = jsonld.flatten(
+            data, {"@context": constants.CONTEXT, "@type": "hydra:Collection"}
+        )
+        versions_graph = _get_single_graph_for_field(data, "@type")
+        if versions_graph is None:
+            return None
+
+        versions_graph["@type"] = "hydra:Collection"
+        # versions_graph["versions"] = versions_graph.pop("versions")
```
mikeAdamss, line 287: you can delete this commented out line

src/store/metadata/oxigraph/store.py
```diff
@@ -306,6 +339,11 @@ def get_version(
             del column["@id"]
         version_graph["table_schema"]["columns"] = columns_graph
         del version_graph["table_schema"]["@id"]
+
+        if type(version_graph["issued"]) != str:
```
mikeAdamss, line 343: you can remove this check and logging, made sense at the time but we've had some issues with data so we're going to push the data validation upstream so its always right when it gets to us, since its right it'll have exactly one issued date.

mikeAdamss (dismissed):
other than a minor comment or two this is completely fine 😄 

if you're still having issues locally then its your local data, try running it with `export GRAPH_DB_URL=https://staging.idpd.uk/sparql` rather than a local oxigraph.

your local oxigraph will cache data so I _suspect_ the duplicate issues were coming from  that rather than anything real, if in doubt (a) docker-compose down then (b) delete the /data folder in your local repo (c). spin it back up and `make populate`.

I'll add a task to make that idempotent as this is not the first time this sort of thing has happened.

NickPapONS (approved):
Looks alright to me, happy the requested changes have been made.

PR #151, opened 2023-11-20
schema.DatasetWithContext added

SarahJohnsonONS:
- `schemas.DatasetWithContext` added, inheriting from `schemas.Dataset`.
- `response_model` in `main.py` `get_all_datasets()` updated to `DatasetWithContext`.
- `schemas.Datasets` updated to return a list of `schemas.Dataset` objects for the `datasets` field. This caused a lot of validation errors, due to some consistency issues with datetimes in the stub content json files. 
- Amended `datasets.json` and all editions/versions json files to be consistent in datetime format used. 
- `schemas.Dataset.spatial_resolution` and `schemas.Dataset.temporal_resolution` - it seems like the `flatten()` method changes a List containing one string to just the string (i.e. not inside a List), so the model now accepts a `Union[str, List[str]]` for these fields.


src/schemas.py
```diff
@@ -162,19 +161,23 @@ class Dataset(BaseModel):
...
+class DatasetWithContext(Dataset):
+    context: str = Field(alias="@context")
+
+
 class Datasets(BaseModel):
     context: str = Field(alias="@context")
     id: str = Field(alias="@id")
     type: List[str] = Field(alias="@type")
-    datasets: List  # TODO - stricter
+    datasets: List[Dataset]  # TODO - stricter
```
mikeAdamss, line 180: remove the todo now its done please.

src/schemas.py
```diff
@@ -162,19 +161,23 @@ class Dataset(BaseModel):
...
     keywords: list[str]
     licence: str
-    spatial_resolution: list[str]
+    spatial_resolution: Union[str, list[str]]
     spatial_coverage: str = Field(pattern=r"^[EJKLMNSW]{1}\d{8}$")
-    temporal_resolution: list[str]
+    temporal_resolution: Union[str, list[str]]
     temporal_coverage: PeriodOfTime
-    editions: List[Union[Edition, SummarisedEdition]]
+    editions: Union[List[SummarisedEdition], List[Edition]]
```
mikeAdamss, line 168: When presented within a Dataset response it'll always be the SummarisedEdition

mikeAdamss (changes requested):
very minor comments only, well done on this

mikeAdamss (approved):
LGTM

PR #158, opened 2023-11-22
Assertions added to data.py

SarahJohnsonONS:
Assertions added to `data.py` to check that the length of the graph increases as resources are added.


nimshi89 (approved):
Seems fine to me

Issue #161, opened 2023-11-22
refactor construct calls


mikeAdamss:
## What is this

These are the construct calls that populate the oxigraph:

https://github.com/GSS-Cogs/idpd-api-poc/blob/main/src/store/metadata/oxigraph/sparql/construct.py

there are two principle problems with them:

(a) we repeat a lot
(b) we'd like an error log where one of them is returning None.

## What to do

Refactor to deal with the two problems above.

I'm not super bothered on implementation, but _something_ like a generic construct method that takes a sparql query would make sense:

```python
condstruct(
     sparl.construct_queries("datasets").format(dataset_id=dataset_id)
     )
```

You will want to make sure `construct` handler any errors resulting from someone not correctly providing the required formatting arguments for a query.

Include what you can in the error log, to 100% include the exact sparql query that returned None.

## Acceptance Criteria

- [ ] Reviewed and merged.
- [ ] We have tests of the above.
- [ ] **one** (to being with) endpoint using the new mechanism.

SarahJohnsonONS:
I've had a look at the SPARQL utilities in csvcubed to see what we can ~~steal~~ borrow. One option is to use `initBindings` to specify the URI of the subject you want to query on - see [this branch] for an idea of how this could be implemented in the `OxigraphMetadataStore.get_dataset` method.

It might be possible to use initBindings for other values in the queries, to reduce duplication - e.g. in `construct_contact_point.sparql` we're querying only on `rdf:type` `dcat:DatasetSeries` - defining a `type` binding would allow the query to be reused for edition contact points as well.

PR #170, opened 2023-11-27
#161 refactor construct calls

SarahJohnsonONS:
- `construct` function defined to take SPARQL `query`, `graph` and any `init_bindings` required to run the query.
  - `logger.info` message logs when queries return no results.
- `OxigraphMetadataStore` methods refactored to use `construct` function.
- `SPARQL_QUERIES` dictionary mapping query name to query text.
- Tests added for `construct` function.
- `get_all_datasets` endpoint updated to use `OxigraphMetadataStore`.


src/store/metadata/sparql.py
```diff
@@ -0,0 +1,230 @@
+from typing import Dict, Optional
+
+from rdflib import Graph
+from rdflib.term import Identifier
+
+from custom_logging import logger
+
+
+def construct(
+    query: str, graph: Graph, init_bindings: Optional[Dict[str, Identifier]] = None
```
mikeAdamss, line 10: where we have init_bindings we'll want to validate them in some way if possible. If the queries borked because of bad bindings its worth us explicitly being told to save us having to unpick sparql.

mikeAdamss (changes requested):
So this works a treat but I'm wary of `dictionary["key"]` patterns in production as it opens the door for typos (and key errors at runtime). So yes to approach but can we refactor it to be a bit more OOP like, something like:

```
class Construct:

    def datasets(self, graph):
       # do /datasets query
       
    def dataset(self, graph, init_bindings):
       # do /datasets/{id} query
       
# Which gets you to

construct = Construct() # Probably at the top somewhere

# So you end up with something like:
        result: Graph = (
            construct.dataset(graph, init_bindings),
            + construct.keywords(graph, init_bindings)
            etc etc
        )
```
Gets you to the same place but you have dot notation so devs cant typo things and know what their options are.

You can _probably_ (have a play) pass self.db into the construct class when you instantiate the store and do a lot of `self.construct.whatever(init_bindings)` calls instead, arguably cleaner again.

src/main.py
```diff
@@ -89,7 +90,7 @@ def get_context(
 def get_all_datasets(
     request: Request,
     response: Response,
-    metadata_store: StubMetadataStore = Depends(StubMetadataStore),
+    metadata_store: OxigraphMetadataStore = Depends(OxigraphMetadataStore),
```
mikeAdamss, line 93: dont swap it over yet, we're still not there

mikeAdamss (commented):
This is excellent, can you just add a test for malformed init bindings (to pair with your new validation) and we'll approve this and get it merged.

mikeAdamss (approved):

PR #173, opened 2023-11-29
Docker collision fixed

Moasib-Arif:


src/store/metadata/oxigraph/store.py
```diff
@@ -41,6 +41,13 @@
 class OxigraphMetadataStore(BaseMetadataStore):
     def setup(self):
         oxigraph_url = os.environ.get("GRAPH_DB_URL", None)
+        if oxigraph_url is None:
```
mikeAdamss, line 44: can see what you're trying to do, but we cant do this, that'd be test logic inside application code we need to keep these things separated.

mikeAdamss (changes requested):
Not sure what the RUNNING_IN_DOCKER stuff is, but you don't need it, it works fine with just the conftest changes 😄 

If you remove the changes in store,py then I'll be happy to approve.

PR #177, opened 2023-12-06
166 time validation

nimshi89:
The method used is the `AwareDatetime` function from pydantic. It passes all the tests and it is very simple to use.


nimshi89:
Ah okay, Yeah I thought when the `Z` was deleted form the time value it did fail the tests so I took it as working. Yeah sure I'll look into it and implement the solution :) Thank you I might take you up on that :D

mikeAdamss:
this is a nice to have and we need to focus on other things so Im going to close the PR. We can pick this up again from the branch in future if we need to.

Issue #164, opened 2023-11-23
enhanced data validation


mikeAdamss:
## What is this

At the moment we're using schemas in data.py to confirm that _each individual_ json document matches the schemas.

However, we're getting issues where dependencies between json files are not respected, we need to validate these dependencies.

## What to do

- [x] If someone specifies an edition @id in the datasets.editions field confirm that edition exists.
- [x] Make sure the modified and issued data for the SummarisedEdition matches the same dates in the corresponding Edition.
- [x] Apply the above two steps to SummarisedVersion and Version from the edition.versions field.
- [x] Where we are referencing topics, make sure those topics exist and have the id that is being referenced.
- [x] Confirm that the "identifier" fields is correctly reflected at the end if the @id url for a given resource.
- [ ] Make sure the "creator" and "publisher" fields are correctly pointing to existing /publisher resources
- anything else you can think of to enforce referential integrity between resources.

### Acceptance Criteria

- [x] data.py is updated to make the validation of submitted resources more dependable.
- [x] (for now), add the `pipenv run python -c "from data import populate;populate(write_to_db=False)"` command (thats what powers "make data") to the github action so malformed data submittions do not pass the test phase. **Put this before the pytest suite** in the action so we know the data is correct and not tripping application code tests for silly reasons.

SarahJohnsonONS:
- `modified` field removed from summarised version object in edition JSON files, as well as `SummarisedVersion` schema model
- OFCOM creator?

PR #178, opened 2023-12-06
#164 enhanced data validation

SarahJohnsonONS:
Validation implemented as specified in the ticket, apart from the creators/publishers part - see L206-210 in `data.py`. Having clarified the issue with Andrew, I'll add a new ticket about how this needs to be resolved, as at the moment, the application profile doesn't align with how publishers and creators are defined in the stub files.

Changes made to JSON files to ensure referential integrity across resources:
- `modified` field removed from `versions` object in edition JSON files, as well as `SummarisedVersion` schema model
- https://staging.idpd.uk/office-for-national-statistics and https://www.gov.uk/government/organisations/office-for-national-statistics replaced with https://staging.idpd.uk/publishers/office-for-national-statistics
- Topics in editions changed to kebab case


data.py
```diff
@@ -35,119 +36,185 @@ def set_context(resource_item):
...
+                    "issued": edn["issued"],
+                    "modified": edn["modified"],
+                }
+            ]
+            versions_in_edition[edn["versions_url"]] = edn["versions"]
+            assert (
+                edn["publisher"] in dataset_publishers
+            ), f"{edn['publisher']} not in publisher list {dataset_publishers}"
+            topics_in_editions.update(edn["topics"])
+        for dataset in datasets_source_dict["datasets"]:
```
NickPapONS, line 150: The nested for loops here might make the function drag on a bit, I personally don't mind but maybe it could benefit from some abstraction/refactoring at this part? But it might also be a bit redundant to just split things up. You know best with this populate function so let me know what you think.

data.py
```diff
@@ -35,119 +36,185 @@ def set_context(resource_item):
...
-        subbed_metadata_store_content_path / "editions/versions"
-    )
+    # Load json files from disk
+    versions = process_json_files(versions_source_path)
 
-    # Validate then add to graph
-    schema = schemas.Versions
-    process_json_files(g, versions_source_path, schema)
+    # Validate data and add to graph
+    for version in versions:
```
NickPapONS, line 166: Possibly more potential for splitting this function up here in the versions, topics and publishers resources.

data.py
```diff
@@ -35,119 +36,185 @@ def set_context(resource_item):
...
     with open(publishers_source_path) as f:
         publishers_source_dict = json.load(f)
-        # Validate then add to graph
-        schemas.Publishers(**publishers_source_dict)
-        g += Graph().parse(
-            data=json.dumps(set_context(publishers_source_dict)),
-            format="json-ld",
-        )
+
+    # TODO This currently fails as OFCOM is not listed as a publisher in any of the datasets (it's a creator for 4gc). Do we need a separate creators.json?
```
NickPapONS, line 206: This failure definitely seems beyond the scope/unrelated to this task so I would probably keep this comment here and consider a new ticket for it.

NickPapONS (commented):
Looks good to me, quite a lot of new code but it is well written and commented. The populate function is definitely long but I can't say that refactoring is strictly necessary, all I would suggest is maybe taking out parts of it into their own functions just to make it shorter, which may not help that much in the end. What do you think? Waiting before giving approval.

PR #179, opened 2023-12-13
Added dedicated test fixtures

Moasib-Arif:
For POC we used stubbed data as test fixtures but to fully move to the graph we need to have dedicated test fixtures.

Changes :
- Copied all the the json-ld into.  the tests/fixtures table
- Created the Oxigraph test fixtures file using the pytest_fixtures mechanisim
- Oxigraph test fixture uses the data.populate function to populate the graph and takes the new location of the Jsonld file as a parameter
- Confest file pass in the jsonld location to data.poulation to spin up a oxigraph for testing


data.py
```diff
@@ -68,8 +68,14 @@ def validate_and_parse_json(g, schema, resource_dict, resource_type):
     g += Graph().parse(data=json.dumps(set_context(resource_dict)), format="json-ld")
     assert len(g) > graph_length
 
-
-def populate(oxigraph_url=None, write_to_db=True):
+# def populate_all_files_in_directory(directory_path, oxigraph_url=None, write_to_db=True):
```
NickPapONS, line 71: Some leftover commented code here

tests/test_dataset.py
```diff
@@ -7,7 +7,7 @@
 from fastapi.exceptions import ResponseValidationError
 
 from constants import JSONLD
-from tests.fixtures.datasets import dataset_test_data
+from tests.fixtures.datasets_oxigraph import dataset_test_data
```
NickPapONS, line 10: Perhaps when the transition to oxigraph is complete and we get rid of all the mentions of the stub, we can rename these files back to not having the "oxigraph" in the name since it would be redundant. But that would come later.

NickPapONS (commented):
Looks good to me, no big changes I have to request so I am happy to look again and give approval when the merge with main is resolved and things pass again. 

The re-implementation of the fixtures in the new directory may lead to a lot of copying over of files/repetition but I think it's fine since it's for the purpose of transitioning over from using the stubs and ideally it will only be temporary.

Moasib-Arif (commented):
Additional Note: In the text fixture JSON data, I included the context for some of the datasets. This was necessary because RDF automatically allocates context to each dataset. Without it, I was encountering a validation error due to a missing context. I'm unsure if this is the correct approach or if there is a better alternative?

mikeAdamss (commented):
We're on the right track here but there's things left to figure out - a good litmus test for this is to delete the contents of `src/store/metadata/stub/content` and try and run the tests. When its done you can run the whole suite of tests without that directory existing.

You'll probably need to look at `StubMetadataStore` as well as its hard coded to use that path for content, so the "live" data is still persisting as part of the stubbed store tests.

You'll need to tweak that class a bit. Have it default to that path but such that we can set the stub store to use test data via (ideally) passing in a new path via a keyword during instantiation or (if that proves tricky) setting a content path via an env var.

There'll be other tweaks no doubt but delete that directory and you should be able to track them down 😄

SarahJohnsonONS (commented):
Happy to approve with just one question - populate() is being run in each of the test fixtures - isn't this already happening in conftest.py? If I comment it out in the fixture, the test still passes.

SarahJohnsonONS (dismissed):
LGTM

SarahJohnsonONS (approved):
Approved

PR #180, opened 2023-12-14
165 log out request

MuazzamChaud:
This ticket makes changes to the custom logger so that 'request_id' is added to it. Then at the start of each request handler, request_id is obtained from the env var 'X-Request-ID' and passed into each store method. Each store method now has a new parameter 'request_id: str' and also have at least one logger.info statement that contains the request_id created earlier, regardless of wether request_id has a value or is none.

How to check:
- Check every request handler gathers request_id from env vars and passes request_id into the store methods.
- check every store methods have new parameter 'request_id'
- check every store method has at least one 'logger.info(...., request_id=request_id)'
- (Optional) See if any more logger statements could be placed anywhere a developer would want to know about whats occurred
- check all logger.info statements in store methods make sense/could do with rewording.
- chcek all unit tests in test_custom_logging.py pass.

### DIFF
--- a/src/custom_logging.py
+++ b/src/custom_logging.py
@@ -1,5 +1,6 @@
 import logging
 import sys
+from typing import Optional
 
 import structlog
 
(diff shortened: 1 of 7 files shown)


NickPapONS (approved):
Looks good! The new logging for the request id seems to be implemented everywhere we need it and the testing is thorough.

Issue #159, opened 2023-11-22
hydrate principle graph


mikeAdamss:
## What is this

So when we use the sparqlstore to pull RDF we then use rdflib to create jsonld.

Most of these responses container sub graphs, so we end up with something (along the lines of, but much bigger than) the following:

```
  {
    "@id": "bnode:bn8d71323061654dd8a5d54b875d733402b115",
    "some_field": "some value",
    "some _other field": "some other value"
  },
  {
     "@id": "https://staging.idpd.uk/the-actual-thing-we-want",
     "a_sub graph": "bnode:bn8d71323061654dd8a5d54b875d733402b115"
  }
```

which _should_ look like:

```
  {
     "@id": "https://staging.idpd.uk/the-actual-thing-we-want",
     "a_sub graph": {
         "some_field": "some value",
         "some _other field": "some other value"
      }
  }
```

Now given the clear relationship between main graph and the sub graphs we can write code to automatically populate the sub documents in the main graph - this is what this task is.

## What to do

This is some code I hacked together that (almost but not quite) does this.

```python
def _hydrate_graph_from_sub_graphs(graph: Dict, sub_graphs: List[Dict]) -> Dict:
    """
    Parses the main graph to find any @id's that are referenes to
    other graphs and populates the sub documents from them.
    """

    def hydrate(data):
        if isinstance(data, dict):
            if "@id" in data.keys() and "@type" not in data.keys():
                data = sub_graphs_dict.get(data["@id"], data)
            else:
                for k, v in data.items():
                    data[k] = hydrate(v)
        elif isinstance(data, list):
            for i, v in enumerate(data):
                data[i] = hydrate(v)
        elif isinstance(data, str):
            data = sub_graphs_dict.get(data, data)
        return data

    sub_graphs_list = [x for x in sub_graphs if x != graph]

    sub_graphs_dict = {
        sub_graph["@id"]:sub_graph for sub_graph in sub_graphs_list
    }

    for sub_graph_dict in sub_graphs_dict.values():
        # Drop any @id fields that signify an anonamous nodes
        if sub_graph_dict["@id"].startswith("bnode:"):
            del sub_graph_dict["@id"]

    graph = hydrate(graph)
    graph["table_schema"] = hydrate(graph["table_schema"])
             
    return graph
```

Now this _should_ work, or at least get you very close if you give it the args (a single graph that is the main graph, a list of graphs which is _all_ the graphs).

Some things to bear in mind:

[...]

PR #181, opened 2023-12-19
#159 hydrate graph

NickPapONS:
Implemented functionality to populate certain fields of the main graph by their relationship with sub graphs. This works by recursively iterating through the contents of the graph and using relationships with ids of the associated sub graphs to "hydrate" the main graph's contents. This notably happens with fields such as the contact_point and the temporal_coverage startDate and endDate.

A test was added that creates a partially implemented graph based off the oxigraph get_dataset function, using the "cpih" dataset as input. But importantly, this deliberately leaves the main graph and sub graphs separate in a list without the embedding/flattening process, to serve as input for the `hydrate` function in the unit test. This leads to some repetition with the oxigraph get_dataset function but I couldn't think of a cleaner way, open to feedback on it.

Edit: the repetition here is now resolved as the test data/json has been added in the form of a fixture.

Initially, two unit tests were made to test each of the two new "hydrate" functions, but I decided that testing the main `hydrate_from_sub_graph` function also implicitly tests the `hydrate` function that is called within it. Also, the assertions within the unit test are checking the contents of the hydrated elements of the resulting modified graph using string comparison. If this isn't what we want I am also happy to try something different.


tests/test_hydrate.py
```diff
@@ -0,0 +1,86 @@
...
+import pytest
+
+from rdflib import Graph, URIRef
+from store.metadata.hydrate import _hydrate_graph_from_sub_graphs
+from pyld import jsonld
+
+from src.store.metadata import constants
+from src.store.metadata.oxigraph.store import OxigraphMetadataStore, _get_single_graph_for_field
+
+def _populate_partial_graph_for_test():
```
mikeAdamss, line 12: this is clever but you can make this a lot easier on yourself.

1.) just run the api and log out (or dump to disk) the results of [this line](https://github.com/GSS-Cogs/idpd-api-poc/blob/44270c961dab0b013358bb6efef631bce31c35e0/src/store/metadata/oxigraph/store.py#L99) and make it a pytest json/dict fixture along the same lines as the other (closer to end product) json fixtures we have. Once you've got that starting point...

2.) use `_get_single_graph_for_field` to get the main graph.

3.) the whole list of graphs should be accessible in that fixture under the `@graph` key.

4.) pass those two things to your function and do your assertions as you already are.

It's not very subtle but it'll be easy to understand which is the main thing really. As long as you've got that fixture as a starting point you can probably just put steps 2 and 3 in the actual test.

mikeAdamss (approved):
LGTM 👍

mikeAdamss:
closing as code was merged on another branch.

PR #183, opened 2023-12-19
Rich additions to the housing stock

rpm-data-dev:
* Made changes based on unit test recommendations including to editions.json for values `issued` and `modified` but the unit test still thinks i'm leaving the old values in there e.g. "2023-05-05T08:59:11.430214" for issued.
* Topics field will need to be updated with values appropriate to the dataset.


src/store/metadata/stub/content/datasets.json
```diff
@@ -318,6 +318,59 @@
...
+      ],
+      "temporal_coverage": {
+        "start": "2022-01-01T00:00:00Z",
+        "end": "2023-12-31T00:00:00Z"
+      },
+      "editions_url": "https://staging.idpd.uk/datasets/additions-to-the-housing-stock/editions",
+      "editions": [
+        {
+          "@id": "https://staging.idpd.uk/datasets/additions-to-the-housing-stock/editions/2022-2023",
+          "issued": "2023-05-05T08:59:11.430214",
```
mikeAdamss, line 370: @rpm-data-dev - this line and the line below is your issue.

you corrected the dates at the edition level, but the gotcha is there's a mini version of the edition in the dataset which needed to be updated too 😄

PR #184, opened 2023-12-22
more validation for data.py

mikeAdamss:
This PR is to accomplish two main goals:

- 1.) adding additional validation to data.py
- 2.) make the validation feedback more obvious so the data engineers can more easily understand what the problem is.

# What I've done

- Added validation of counts, so that if the count field in a json does not match the number of sub documents an assertion error is raised.
- Added explicit messages where the short form documents of editions and versions have field values that do not match the field values in the long form representation of same.
- Validate that in_series is specifying something sensible.
- Moved data.py into /src (it needs to be here as we'll need to add tests soon as its getting complicated).
- Updated the github action and makefile to call data.py in its new location.
- Confirms that all creators and publishers referenced in datasets actually exist, and that not publishers are being created that are not referenced via a dataset.

# How to review

It's getting to be quite a complex file and its hard to see the changes as I've moved the file, a line by line review will not be necessary so just sanity check the functionality.

The easiest way to sanity check will be to modify the stubbed data to be wrong (misalign an issued date etc, things from the above list basically) and `make data` to confirm a suitable message is triggered.


NickPapONS (approved):
LGTM, no changes to request 👍

PR #185, opened 2023-12-27
remote oxigraph population instructions

mikeAdamss:
These are the instructions to allow someone to sync the remote oxigraph server such that it has all the RDF to represent the data being served via the api.

## How to review

Follow the instructions, make sure it works as advertised.

**Note** - you'll need some platform related command I can't put in a public github document. Please check slack or put a message in the team help channel.

The main feedback we're after is whether or not this tells you everythign you need to know, if you have to work bits out yourself then while that's good, it'd be nice to know so we can make it clearer for the next person.


mikeAdamss:
closing as no longer relevant

PR #195, opened 2024-01-05
disable all tests using oxigraph

mikeAdamss:
Curren the stubbed data for the api is also used to populate a docker ran oxigraph we use for testing oxigraph based functionality.

Now that more data is being added, this is leading to http timeouts during the graph population process.

Given that moving to dedicated test fixtures is already in flight, this PR is just to disable the oxigraph tests (as we're not using oxigraph anyway) as these failures are currently blocking the DE pr's.

## How to review

`make test`

make sure it works, make sure no oxigraph tests are being ran. sanity check

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,54 +1,54 @@
-import time
+# import time
 
-import docker
-from docker import DockerClient
-import pytest
+# import docker
+# from docker import DockerClient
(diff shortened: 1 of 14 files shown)


NickPapONS (approved):
All tests pass, agree with ones commented out 👍

Issue #190, opened 2024-01-02
validation - sub edition and sub version


mikeAdamss:
## What is this

Inside each datasets metadata there is a "datasets" field with a list of truncated dataset sub documents: [This is an example](https://github.com/GSS-Cogs/idpd-api-poc/blob/feab65693ec350e5b03af3b80d7b9ff7bcb91e61/src/store/metadata/stub/content/datasets.json#L54).

Inside each editions metadata there is a "versions" field with a list of truncated version sub documents. [This is an example](https://github.com/GSS-Cogs/idpd-api-poc/blob/5c2861af9a694d699dc9b35d67e6ec97d2f295af/src/store/metadata/stub/content/editions/4gc_2023-09.json#L50).

Both of these fields _should be ordered_ by the `issued` field, starting with the most recent. This task is to add validation to the load script (data.py) to confirm this.

## What to do

- Add validation to data.py to confirm the ordering in all "editions" and "versions" fields - throw an error if the ordering is wrong. Make sure the error is obvious and explains the issue in plain english.

## Acceptance Criteria

- [ ] Validation Added.
- [ ] Add a simple unit test to confirm that this validation is working as intended.

PR #198, opened 2024-01-09
made the method and test

nimshi89:
#190
This is for a method to validate 'editions' and 'versions' subtopic are ordered by 'issued'.


src/data.py
```diff
@@ -395,6 +398,46 @@ def _assert_summarised_edition_in_dataset(datasets: Dict, edition: Dict):
...
 
+def _assert_editions__ordered_by_issued(list_of_datasets, topic: str, sub_topic: str):
+    """this function will assert that the list of datasets
+     is ordered by issued, raises error if not"""
+
+    index  = 0
+
+    list_of_issued=[]
+    
+    for x in list_of_datasets[topic]:
```
mikeAdamss, line 411: try and use descriptive variables please, "x" is only really acceptable inside a comprehension (its an _anonymous_ function). So something more like`for dataset_dict in list_of_datasets` etc, always think of what favours you can do for the next programmer to read your code.

mikeAdamss (commented):
given that in both cases you're checking the issued field on a list of dictionaries I was envisioning something like:

```
def assert_ordered_by_issued_field(list_of_dicts: List[Dict):
       # Some logic to confirm the list of dictionaries is in
       # the right order according to the issued field.
       # see comments below this example
       
def _assert_editions_ordered_by_issued(datasets_json_as_dict)
       list_of_editions_dicts = datasets_json_as_dict["editions"]
       assert_ordered_by_issued_field(list_of_editions_dicts)
       
def _assert_versions_ordered_by_issued(editions_json_as_dict)
       list_of_versions_dicts = editions_json_as_dict["versions"]
       assert_ordered_by_issued_field(list_of_versions_dicts)
```

You can probably simplify the `assert_ordered_by_issued_field` logic, _something_ along the lines of:

- 1.) get the issued value from the first dict in the list of dicts and convert it to time.
- 2.) iterate the list of dicts converting each issued value to time and asset that its not a later time than you set in 1

that ought to work and be relatively concise.

mikeAdamss:
closing as code is now part of a different pr.

PR #212, opened 2024-01-19
#202 extract validation from data.py

SarahJohnsonONS:
Validation functionality moved to `src/store/metadata/stub/validation.py`, and associated tests added in `/tests/store/metadata/stub/test_validation.py`, using test cases in `tests/store/metadata/stub/test-cases`.

Added `make validation` to Makefile and updated the Github Action to validate the data before populating the store. Added assertions for length of graph back into `populate()`.


src/data.py
```diff
@@ -245,213 +158,53 @@ def skolemise(node):
...
-    1. A short form representation of a version is included in a "versions" sub
-       document within the edition json file for the relevant edition.
-    2. The full version documents is included in the versions json file.
-
-    With this check we are making sure that the field values as inlcuded in the
-    short form version match the fields as included in the full version document.
-    """
-
-    for id, issued in [(x["@id"],x["issued"]) for x in version["versions"]]:
+# def _assert_editions_ordered_by_issued(list_of_datasets, main_dict: str, sub_dict: str):
```
mikeAdamss, line 161: can we get rid of. the commented our code in this file please

Makefile
```diff
@@ -14,6 +14,9 @@ lint: ## Run the ruff python linter (auto triggered on pre-commit)
 test: ## Run pytest and check test coverage (auto triggered on pre-push)
 	pipenv run pytest --cov-report term-missing --cov=src --cov-config=./tests/coverage.rc
 
+validate: ## Validate stub data files
+	pipenv run python3 ./src/store/metadata/stub/validation.py
+
 populate: ## Populate an oxigraph DB (that/if/when) one is running on localhost:7878
 	pipenv run python3 ./src/data.py
```
mikeAdamss, line 21: can we add the validate make command to run prior to this one (chaining make commands is a thing you can do). Lets 100% make sure nobody skips a step and tries to populate the graph with incorrect data.

src/store/metadata/stub/validation.py
```diff
@@ -0,0 +1,423 @@
...
+        IDs of publisher resources being used:
+        {unique_publishers_referenced}
+        """
+        )
+
+    # Schema validation
+    schemas.Publishers(**publishers)
+
+
+def _assert_editions_ordered_by_issued(list_of_datasets, main_dict: str, sub_dict: str):
```
mikeAdamss, line 329: I dont think this is getting called anywhere. If I'm following this should be getting called in `validate_datasets()`

src/store/metadata/stub/validation.py
```diff
@@ -0,0 +1,423 @@
...
+                list_of_issued.append(fornow)
+                index += 1
+
+            most_recent = max(list_of_issued)
+            assert (
+                first_value == most_recent
+            ), "The datasets should be ordered by 'issued' (from most recent)"
+
+
+def _assert_versions_ordered_by_issued(datasets, sub_dict: str):
```
mikeAdamss, line 354: don't _think_ this is getting called either, think it needs to be called somewhere in `validate_editions()`

mikeAdamss (commented):
this all looks sensible to me, very minor comments only.

one loose thread - we've for json test fixtures in both:
- `/tests/metadata/stub/test-cases`
- `/tests/test_jsons`

it would be better if they were all in the former (but gven the context, if that's particularly arduous we'll live with it as-is).

src/store/metadata/stub/validation.py
SarahJohnsonONS: Will delete - I rewrote it and added it to _validate_version_in_edition L175-L184

src/store/metadata/stub/validation.py
SarahJohnsonONS: Will delete - I rewrote it in _validate_edition_in_dataset L100-L109

Makefile
SarahJohnsonONS: I think this works...

mikeAdamss (approved):
LGTM 👍

PR #224, opened 2024-01-29
191 request csv

Moasib-Arif:
Request for csv to /datasets/{id} or datasets/{id]/editions/{id} -so it returns the most recent csv versions

Changes:
- Added Requests to get mimetype text/csv for /datasets/{id} and datasets/{id}/editions/{id} so it returns the latest versions from CSVStub Store and added tests

Note:
Assumpution - that the items in the "editions" and "versions" lists are ordered with the most recent at the top ( Ticket [190](https://github.com/GSS-Cogs/idpd-api-poc/issues/190))


src/main.py
```diff
@@ -147,7 +148,17 @@ def get_dataset_by_id(
     request_id = request.headers.get("X-Request-ID", None)
     logger.info("Received request for dataset with ID", data={"dataset_id": dataset_id}, request_id=request_id)
 
-    if request.headers["Accept"] == JSONLD or BROWSABLE:
+    if request.headers["Accept"] == CSV:
```
mikeAdamss, line 151: Can you add a comment above this to explain what it is. eg "If someone is asking for a csv at the dataset level we return the latest version of the latest edition.

If we come back and read this at a future date that'll be a really handy thing to know.

src/main.py
```diff
@@ -240,7 +252,16 @@ def get_dataset_edition_by_id(
         request_id=request_id,
     )
 
-    if request.headers["Accept"] == JSONLD or BROWSABLE:
+    if request.headers["Accept"] == CSV:
```
mikeAdamss, line 255: Can you add a comment above this to explain what it is. eg "If someone is asking for a csv at the edition level we return the latest version of the this edition.

If we come back and read this at a future date that'll be a really handy thing to know.

src/store/csv/stub/store.py
```diff
@@ -58,3 +93,26 @@ def get_version(self, dataset_id: str, edition_id: str, version_id: str, request
         )
 
         return FileResponse(csv_path) if csv_path else None
+
+
+    def get_latest_version(self, csv_identifier: str) -> Optional[str]:
```
mikeAdamss, line 98: there's a nuance here I probably didn't capture on the card.

- if someone requests a csv at the dataset level we return the latest verion of the **latest** edition.
- if someone requests a csv at the edition level we return the latest version of **that** edition.

haven't dug into this code properly so if its already doing that then ignore this comment, if its not can you change it to do so please.

mikeAdamss (commented):
code is excellent, some minor points only.

mikeAdamss (approved):
LGTM

PR #227, opened 2024-02-06
API bug fixed

SarahJohnsonONS:
**kwargs in BaseMetadataStore.__init__ was causing a Pydantic validation error - removed it and added StubMetadataStore.__init__ method.


NickPapONS (approved):
Looks good to me, all test pass 👍

